PR #12, opened 2024-01-31
Add kairos nixos module

marijanp:
This PR introduces:
- a nixos-module for the kairos-server
- a kairos-host nixos-configuration
- a nixos-test verifying that the nixos-module and the configuration work as expected
- a basic e2e nixos-test

### DIFF
--- a/nixos/modules/kairos.nix
+++ b/nixos/modules/kairos.nix
@@ -1,0 +1,68 @@
+{ lib, pkgs, config, ... }:
+let
+  inherit (lib)
+    types
+    mkOption
+    mkIf
+    mkMerge
+    mkEnableOption
+    mdDoc
+    ;
+  cfg = config.services.kairos;
+in
+{
+  options.services.kairos = {
+
+    enable = mkEnableOption (mdDoc "kairos");
+
+    package = mkOption {
+      type = types.package;
+    };
+
+    port = mkOption {
+      type = types.port;
+      default = 60000;
+      example = 60000;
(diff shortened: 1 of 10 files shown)


nixos/modules/kairos.nix
```diff
@@ -0,0 +1,68 @@
...
+        "info"
+        "debug"
+        "trace"
+      ];
+      default = "info";
+      description = ''
+        The log-level that should be used.
+      '';
+    };
+  };
```
Avi-D-coder, line 44: This should have a string variant for supporting more complex log levels `RUST_LOG=[target][=][level][,...]`
https://docs.rs/env_logger/0.11.1/env_logger/#enabling-logging

Avi-D-coder (approved):

nixos/modules/kairos.nix
marijanp: Since we don't need more sophisticated logging configuration yet, I would leave it as is to avoid spending the amount of work needed to replicate this is in nix code.

koxu1996:
**Comment:** I am not convinced that NixOS VM is the best way to test project in CI/CD, as - I guess - it takes a lot of resources, and needs lot of additional host/network/user configuration.

Overall, PR looks good and can be merged :wink:.

marijanp:
The configuration you see in this PR is a configuration for a server that could be deployed into production. I use the same configuration inside the VM test. This means, the e2e test not only tests the application but also the configuration that we can use in production.

Issue #14, opened 2024-02-01
Implement CLI parsing


koxu1996:
We need 3 basic commands: deposit / transfer / withdraw.

Issue #7, opened 2024-01-24
Verify signatures


Avi-D-coder:
Verify signatures where I have TODOs on transactions.
Favor RustCrypto k256 so we can match the use of the risc0 accelerated patch in the guest.
https://github.com/risc0/risc0/tree/release-0.19/examples/ecdsa

PR #17, opened 2024-02-05
Base CLI parsing

koxu1996:
## CLI interface

Implemented interface with three key transaction commands (Deposit, Transfer, Withdraw) and integrated Casper's cryptography functionalities.

Fixes https://github.com/cspr-rad/kairos/issues/14.

**Deposit:**

```sh
$ ./kairos-cli deposit \
  --amount 123 \
  --private-key ~/wallet/secret_key.pem
```

**Transfer:**

```sh
$ ./kairos-cli transfer \
  --recipient 01a26419... \
  --amount 123 \
  --private-key ~/wallet/secret_key.pem
```

**Withdraw:**

```sh
$ ./kairos-cli withdraw \
  --amount 123 \
  --private-key ~/wallet/secret_key.pem
```

### Outputs

- amount: `u64` - no need to handle u512 at this point
- signer: `CasperSigner` (created from `CasperPrivateKey`)
- recipient: `CasperPublicKey`

Notably, signer allows generating signature of arbitrary byte data. It will be used to sign transactions sent to Kairos server.

### Error handling

1) Invalid transaction amount, for example `123a`:

```
error: invalid value '123a' for '--amount <NUM_MOTES>': invalid digit found in string
```

2) Invalid recipient `foobar`, when making transfer:

```
error: invalid value 'foobar' for '--recipient <PUBLIC_KEY>': failed to parse hex string: Invalid character 'o' at position 1
```

3) Invalid private key path, when content cannot be read:

```
cryptography error: failed to parse private key: secret key load failed: could not read '/non/exisiting/file/path': No such file or directory (os error 2)
```

4) Invalid private key content:

```
cryptography error: failed to parse private key: pem error: malformedframing
```

### Tests

Both 3 happy paths and error cases are tested in `cli_tests.rs`.


Avi-D-coder (changes requested):

kairos-cli/bin/main.rs
```diff
@@ -1,3 +1,46 @@
...
+            process::exit(1);
+        }
+    };
+
+    match result {
+        Ok(output) => {
+            println!("{}", output)
+        }
+        Err(error) => {
+            println!("{}", error);
```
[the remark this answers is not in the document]
koxu1996: Fixed in https://github.com/cspr-rad/kairos/pull/17/commits/0dafb6487f3981b0293cfa3163bb46bfc377a39e.

kairos-cli/bin/commands/mod.rs
```diff
@@ -0,0 +1,24 @@
...
+pub trait ClientCommand {
+    const NAME: &'static str;
+    const ABOUT: &'static str;
+
+    /// Constructs the clap subcommand.
+    fn new_cmd() -> Command;
+
+    /// Parses the arg matches and runs the subcommand.
+    fn run(matches: &ArgMatches) -> Result<Output, CliError>;
+}
```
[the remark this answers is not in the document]
koxu1996: Ok, removed in https://github.com/cspr-rad/kairos/pull/17/commits/c63d8d3dfa777b9e4c6f85599f5b46a55251f0e5.

kairos-cli/tests/cli_tests.rs
```diff
@@ -0,0 +1,100 @@
...
+use std::path::PathBuf;
+
+// Helper function to get the path to a fixture file
+fn fixture_path(relative_path: &str) -> String {
+    let mut path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+    path.push("tests");
+    path.push("fixtures");
+    path.push(relative_path);
+    path.to_str().unwrap().to_string()
+}
```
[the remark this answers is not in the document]
koxu1996: Updated in https://github.com/cspr-rad/kairos/pull/17/commits/65c3eb3f3938c80132c445886e5efbff294e792a.

koxu1996:
Solved issue with failing Nix tests: https://github.com/cspr-rad/kairos/pull/17/commits/ade64d17d0336cc089f7b7077d89d3c5b0725589 :muscle:.

kairos-cli/bin/commands/deposit.rs
```diff
@@ -0,0 +1,31 @@
+use crate::common::{amount, private_key};
+use crate::crypto::signer::CasperSigner;
+use crate::error::CliError;
+use clap::{ArgMatches, Command};
+
+pub struct Deposit;
```
marijanp, line 6: Please make this more declarative to reduce the cognitive overhead for finding out what arguments the deposit command expects and what the arguments types are i.e. use claps derive features https://docs.rs/clap/latest/clap/_derive/index.html
```rust
#[derive(Parser)]
pub struct Cli {
    #[clap(subcommand)]
    pub command: Command,
}
#[derive(Subcommand)]
pub enum Command {
    #[clap(subcommand)]
    Deposit {
    
    },
    Transfer {
    },
    Withdraw {
    }
}
#[derive(Subcommand)]
pub enum ProveCommand {
    Hypotenuse {
        
        a: u64,
        #[clap(short, long)]
        b: u64,
    },
    Polynomial {
        #[clap(short, long)]
        a: i64,
        #[clap(short, long)]
        x: i64,
        #[clap(short, long)]
        b: i64,
    },
}
```
In order to find out what arguments the `Deposit` command requires I have to first read `new_cmd` then I have to look for `amount::arg()` in a completely different file. In that file I then have to look up static variables to figure out their names are and in order to find the type or the argument I have to read the `get()` method.

kairos-cli/bin/crypto/error.rs
```diff
@@ -0,0 +1,10 @@
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum CryptoError {
+    /// Failed to parse a public key from a formatted string.
```
marijanp, line 5: This comment is not necessary as it repeats the code

kairos-cli/bin/crypto/private_key.rs
```diff
@@ -0,0 +1,15 @@
+use crate::crypto::error::CryptoError;
+
+pub struct CasperPrivateKey(pub casper_types::SecretKey);
+
+impl CasperPrivateKey {
+    pub fn from_key(public_key: casper_types::SecretKey) -> Self {
```
marijanp, line 6: this method is shallow, i.e. what is the difference between calling
```
let casper_key = CasperPrivateKey(public_key);
// and
let casper_key = CasperPrivateKey::from_key(public_key);
// then there could also be
let casper_key = CasperPrivateKey::new(public_key);
```
but the main issue is that this method doesn't do anything.

marijanp: If you compare it with `from_file`: `from_file` provides some funcitonality

kairos-cli/bin/crypto/public_key.rs
```diff
@@ -0,0 +1,28 @@
...
+
+        Ok(Self(public_key))
+    }
+
+    pub fn from_key(public_key: casper_types::PublicKey) -> Self {
+        Self(public_key)
+    }
+
+    #[allow(unused)]
+    fn get(&self) -> Result<Vec<u8>, CryptoError> {
```
marijanp, line 23: this should be called `to_bytes` or `as_bytes`

kairos-cli/bin/error.rs
```diff
@@ -0,0 +1,15 @@
...
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum CliError {
+    /// Unable to find argument by name.
+    #[error("missing argument '{context}'")]
+    MissingArgument { context: &'static str },
+    /// Failed to parse amount from string.
+    #[error("failed to parse '{context}' as u64")]
+    FailedToParseU64 { context: &'static str },
```
marijanp, line 11: these are errors that clap has built in

kairos-cli/bin/main.rs
```diff
@@ -1,3 +1,46 @@
...
+use clap::Command;
+use commands::{Deposit, Transfer, Withdraw};
+use std::process;
+
+fn cli() -> Command {
+    Command::new("Kairos Client")
+        .about("CLI for interacting with Kairos")
+        .subcommand(Deposit::new_cmd())
+        .subcommand(Transfer::new_cmd())
+        .subcommand(Withdraw::new_cmd())
```
marijanp, line 15: Imagine someone adding a new subcommand, not being aware that this needs to be extended.

kairos-cli/bin/main.rs
```diff
@@ -1,3 +1,46 @@
...
+    let (subcommand_name, matches) = arg_matches.subcommand().unwrap_or_else(|| {
+        // No subcommand provided by user.
+        let _ = cli().print_long_help();
+        process::exit(1);
+    });
+
+    let result = match subcommand_name {
+        Deposit::NAME => Deposit::run(matches),
+        Transfer::NAME => Transfer::run(matches),
+        Withdraw::NAME => Withdraw::run(matches),
```
marijanp, line 29: If we had a typed CLI nobody would ever forget to handle a subcommand because the compiler would warn the developer that there is a unmatched case

kairos-cli/tests/cli_tests.rs
```diff
@@ -0,0 +1,98 @@
...
+#[test]
+fn deposit_successful_with_ed25519() {
+    let secret_key_path = fixture_path("ed25519/secret_key.pem");
+
+    let mut cmd = Command::cargo_bin("kairos-cli").unwrap();
+    cmd.arg("deposit")
+        .arg("--amount")
+        .arg("123")
+        .arg("--private-key")
+        .arg(secret_key_path);
```
marijanp, line 20: there is also `cmd.args(["deposit", "--amount", "123", "--private-key", secret_key_path])`

kairos-cli/Cargo.toml
```diff
@@ -13,3 +13,8 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+assert_cmd = "2.0.13"
```
marijanp, line 16: this can be added to `[dev-dependencies]` as it is a dependency used in tests only. I'm also not convinced that this dependency is necessary as
one could simply use assert_eq to verify that the output of a command matches, moreover the `cargo_bin("kairos-cli")` can be substituted with `Command::new(env!("CARGO_BIN_EXE_kairos-cli"))`

marijanp (commented):
I would like to see code that leverages the type system and is more more declarative. That leaves us with code that is easier to follow and extend. Moreover, the compiler supports us to catch missing/ unhandled commands.

kairos-cli/bin/crypto/error.rs
koxu1996: This is intended for documentation:

image

kairos-cli/bin/crypto/private_key.rs
koxu1996: This was abstraction leftover, cleaned in https://github.com/cspr-rad/kairos/pull/17/commits/0744fd55493fc154bb0f2bbd9d608d922501a5b6.

kairos-cli/bin/crypto/public_key.rs
koxu1996: Ok, renamed in https://github.com/cspr-rad/kairos/pull/17/commits/d6a9135ba7086c67ae58aca65cb7302fa2ba8131.

kairos-cli/tests/cli_tests.rs
koxu1996: This would make code more complex, as `args()` works only with `&str`, so passing `secret_key_path` requires additional work to convert it from `&Path` into `&str` (+handle error).

kairos-cli/Cargo.toml
koxu1996: Moved in https://github.com/cspr-rad/kairos/pull/17/commits/fd4476e05a88298df87fb7eb9ca3f14f1bf3c859.

kairos-cli/bin/main.rs
```diff
@@ -1,3 +1,46 @@
...
+}
+
 fn main() {
-    println!("Hello, world!");
+    let arg_matches = cli().get_matches();
+    let (subcommand_name, matches) = arg_matches.subcommand().unwrap_or_else(|| {
+        // No subcommand provided by user.
+        let _ = cli().print_long_help();
+        process::exit(1);
+    });
```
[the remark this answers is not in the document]
koxu1996: Matching subcommand requires unwrapping `arg_matches`, no matter if we build CLI with derive macros or not.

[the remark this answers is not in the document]
marijanp: You would just call `clap::parse` and clap takes care of handling parse errors and exiting.
```
let cli = Cli::parse();
match &cli.command {
```
where `Cli` is the type that describes the interface

kairos-cli/bin/commands/deposit.rs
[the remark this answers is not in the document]
koxu1996: Improved in https://github.com/cspr-rad/kairos/pull/17/commits/f54366613e48e8b5f9f5410a26b2b43124965881:

```rs
#[derive(Parser)]
#[command(name = "Kairos Client", about = "CLI for interacting with Kairos")]
struct Cli {
    #[command(subcommand)]
    command: Command,
}

...

#[derive(Subcommand)]
pub enum Command {
    #[command(about = "Deposits funds into your account")]
    Deposit(deposit::Args),
    #[command(about = "Transfers funds to another account")]
    Transfer(transfer::Args),
    #[command(about = "Withdraws funds from your account")]
    Withdraw(withdraw::Args),
}

...

#[derive(Parser)]
pub struct Args {
    #[arg(long, short, value_name = "PUBLIC_KEY", value_parser = parse_hex_string)]
    recipient: StdVec<u8>,
    #[clap(flatten)]
    amount: AmountArg,
    #[clap(flatten)]
    private_key_path: PrivateKeyPathArg,
}

...


#[derive(Args, Debug)]
pub struct AmountArg {
    #[arg(name = "amount", long, short, value_name = "NUM_MOTES")]
    pub field: u64,
}

...
```

kairos-cli/bin/main.rs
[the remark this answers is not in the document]
koxu1996: I was unaware of it :sweat_smile: Already fixed in https://github.com/cspr-rad/kairos/pull/17/commits/f54366613e48e8b5f9f5410a26b2b43124965881:

```rust
    let cli = Cli::parse();

    let result = match cli.command {
        Command::Deposit(args) => commands::deposit::run(args),
        Command::Transfer(args) => commands::transfer::run(args),
        Command::Withdraw(args) => commands::withdraw::run(args),
    };
```

kairos-cli/bin/error.rs
koxu1996: Error handling improved in https://github.com/cspr-rad/kairos/pull/17/commits/f54366613e48e8b5f9f5410a26b2b43124965881, now those error are handled by `clap`.

kairos-cli/bin/main.rs
koxu1996: This got fixed with refactor: https://github.com/cspr-rad/kairos/pull/17/commits/f54366613e48e8b5f9f5410a26b2b43124965881.

kairos-cli/bin/main.rs
koxu1996: This got fixed with refactor: https://github.com/cspr-rad/kairos/pull/17/commits/f54366613e48e8b5f9f5410a26b2b43124965881.

In case of unmatched case, the following error is thrown:

```
error[E0004]: non-exhaustive patterns: `commands::Command::Foo(_)` not covered
```

koxu1996:
Ready for re-review :eyes:.

kairos-cli/bin/commands/transfer.rs
```diff
@@ -0,0 +1,31 @@
+use crate::common::args::{AmountArg, PrivateKeyPathArg};
+use crate::crypto::public_key::CasperPublicKey;
+use crate::crypto::signer::CasperSigner;
+use crate::error::CliError;
+use crate::utils::parse_hex_string;
+
+use clap::Parser;
+
+type StdVec<E> = Vec<E>;
```
Avi-D-coder, line 9: What's this alias for?

kairos-cli/bin/crypto/signer.rs
```diff
@@ -0,0 +1,54 @@
+use std::path::PathBuf;
+
+use super::private_key::{parse_private_key, CasperPrivateKey};
+use super::public_key::CasperPublicKey;
+use crate::crypto::error::CryptoError;
+use casper_types::{bytesrepr::ToBytes, SecretKey};
+use casper_types::{crypto, PublicKey};
```
Avi-D-coder, line 7: Nit: merge these imports.
RLS has an action for this.

kairos-cli/bin/crypto/signer.rs
```diff
@@ -0,0 +1,54 @@
...
+        // Derive the public key.
+        let public_key = CasperPublicKey::from_key(PublicKey::from(&secret_key.0));
+
+        CasperSigner {
+            secret_key,
+            public_key,
+        }
+    }
+
+    pub fn from_file(secret_key_path: &str) -> Result<Self, CryptoError> {
```
Avi-D-coder, line 26: This and `from_key_pathbuf` are duplicates.
Call `crate::crypto::private_key::CasperPrivateKey::from_file`  instead of calling `casper_types::from_file`.
If we have our own `CasperPrivateKey` newtype we should use it consistently.

kairos-cli/bin/crypto/error.rs
```diff
@@ -0,0 +1,14 @@
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum CryptoError {
+    /// Unable to load a file from the given path.
+    #[error("failed to load key from file")]
+    KeyLoad,
+    /// Failed to parse a public key from a raw data.
+    #[error("failed to parse private key")]
+    FailedToParseKey,
```
Avi-D-coder, line 10: I don't love `casper_types::crypto::error::ErrorExt`, but we should use it anyway.

kairos-cli/bin/crypto/private_key.rs
```diff
@@ -0,0 +1,18 @@
...
+
+pub struct CasperPrivateKey(pub casper_types::SecretKey);
+
+impl CasperPrivateKey {
+    pub fn from_file(file_path: &str) -> Result<Self, CryptoError> {
+        let data = read_file(file_path).map_err(|_e| CryptoError::KeyLoad)?;
+        let secret_key =
+            casper_types::SecretKey::from_pem(data).map_err(|_e| CryptoError::FailedToParseKey)?;
+        Ok(Self(secret_key))
+    }
```
Avi-D-coder, line 13: Change the arguments to match their signature. It use `&Path` via `AsRef`, using `&Path` directly is fine as well.

```suggestion
    pub fn from_file<P: AsRef<Path>>(file_path: P) -> Result<Self, ErrorExt> {
        casper_types::SecretKey::from_file(file_path).map(Self)
    }
```

Avi-D-coder (changes requested):

kairos-cli/bin/commands/transfer.rs
marijanp: This is not needed in my opinion, especially given that it does not instantiate the generic type. It does not make the code more obvious.

marijanp: It could instead be `type Bytes = std::Vec<u8>`

kairos-cli/bin/commands/mod.rs
```diff
@@ -0,0 +1,15 @@
+pub mod deposit;
+pub mod transfer;
+pub mod withdraw;
+
+use clap::Subcommand;
+
+#[derive(Subcommand)]
+pub enum Command {
+    #[command(about = "Deposits funds into your account")]
+    Deposit(deposit::Args),
```
marijanp, line 10: Factoring out these into separate modules increases obscurity. it requires a reader to go to the module and figure out what args the deposit command requires. I'm taking the perspective of someone just wanting to know what our CLI is able to do. I would much more prefer to see that in one file.

I'm also taking into account that you are trying to create a pattern where every command has a `run` function and it solely operates on respective arguments, which I think is great.

But I would suggest creating a `cli` module putting the CLI types in it together with the respective args. This would obviously cause some name collisions for the `Args` type. Args is a very generic name, instead I would suggest introducing `DepositArgs`, `TransferArgs`, and `WithdrawArgs`.

These type names make it very obvious about which arguments we are actually talking and by putting the whole CLI description into one file, someone new to the project only needs to read one module to know what it's capable to do.

kairos-cli/bin/crypto/error.rs
marijanp: Do you mean it's missing here?

kairos-cli/bin/commands/transfer.rs
marijanp: But i would actually say remove it because you need to alias &vec<u8> too then I assume

kairos-cli/bin/crypto/private_key.rs
```diff
@@ -0,0 +1,18 @@
...
+}
+
+pub fn parse_private_key(file_path: &str) -> Result<CasperPrivateKey, CryptoError> {
+    CasperPrivateKey::from_file(file_path)
+}
```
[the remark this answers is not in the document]
marijanp: It's moreover not really parsing it in this function

kairos-cli/bin/crypto/public_key.rs
```diff
@@ -0,0 +1,29 @@
...
+        let (public_key, _remainder) =
+            casper_types::PublicKey::from_bytes(bytes).map_err(|_e| {
+                CryptoError::Serialization {
+                    context: "public key",
+                }
+            })?;
+        Ok(Self(public_key))
+    }
+
+    pub fn from_key(public_key: casper_types::PublicKey) -> Self {
```
marijanp, line 19: I don't think there is a benefit having this function, users would simply pattern-match to get the inner value

kairos-cli/bin/crypto/signer.rs
```diff
@@ -0,0 +1,54 @@
...
+    pub fn from_key_pathbuf(secret_key_path: PathBuf) -> Result<Self, CryptoError> {
+        let private_key_path_str: &str = secret_key_path.to_str().ok_or(CryptoError::KeyLoad)?;
+        let private_key = parse_private_key(private_key_path_str)?;
+
+        Ok(Self::from_key_raw(private_key))
+    }
+
+    pub fn get_public_key(&self) -> CasperPublicKey {
+        self.public_key.clone()
+    }
```
marijanp, line 42: This hides a clone, just let the user access it throught the publicly exposed attribute and let the user decide what to do.

kairos-cli/bin/common/args.rs
```diff
@@ -0,0 +1,15 @@
...
+
+#[derive(Args, Debug)]
+pub struct AmountArg {
+    #[arg(name = "amount", long, short, value_name = "NUM_MOTES")]
+    pub field: u64,
+}
+
+#[derive(Args, Debug)]
+pub struct PrivateKeyPathArg {
+    #[arg(name = "private-key", long, short = 'k', value_name = "FILE_PATH")]
```
marijanp, line 13: can we use value_parse here to immediately parse this to a private-key?

marijanp: An interesting read on this btw: https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/

kairos-cli/bin/error.rs
```diff
@@ -0,0 +1,21 @@
...
+
+#[derive(Error, Debug)]
+pub enum CliError {
+    /// Cryptography error.
+    #[error("cryptography error: {error}")]
+    CryptoError {
+        #[from]
+        error: CryptoError,
+    },
+    // TODO: Add error for "Failed to parse hex string: {}"
```
marijanp, line 14: can this todo be removed?

marijanp (commented):
Great, I think we are almost there. One last remark. I've put the `main.rs` into build to create a separation between library and binary. If you could put all the modules except `main.rs` into a `src` directory that would be perfect.

kairos-cli/bin/crypto/private_key.rs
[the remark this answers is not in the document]
koxu1996: Removed in https://github.com/cspr-rad/kairos/pull/17/commits/5dcfe2819ce6c7b83ec1ff078fc5d6999b3ef864.

kairos-cli/bin/crypto/error.rs
Avi-D-coder: Yeah, it should be wrapped in a variant here.

kairos-cli/bin/crypto/signer.rs
koxu1996: Imports formatted: https://github.com/cspr-rad/kairos/pull/17/commits/c314c50b33c09e6f3e9449bc9d80b3d9fca63fca.

kairos-cli/bin/crypto/signer.rs
```diff
@@ -0,0 +1,54 @@
...
+
+    pub fn from_file(secret_key_path: &str) -> Result<Self, CryptoError> {
+        let secret_key =
+            SecretKey::from_file(secret_key_path).map_err(|_e| CryptoError::FailedToParseKey)?;
+
+        Ok(Self::from_key_raw(CasperPrivateKey(secret_key)))
+    }
+
+    pub fn from_key_pathbuf(secret_key_path: PathBuf) -> Result<Self, CryptoError> {
```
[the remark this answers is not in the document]
koxu1996: This duplicate function was removed in https://github.com/cspr-rad/kairos/pull/17/commits/b1a78b47dfce7f834e6d5f05e824b217beffa53f.

kairos-cli/bin/crypto/signer.rs
koxu1996: Fixed in https://github.com/cspr-rad/kairos/pull/17/commits/b1a78b47dfce7f834e6d5f05e824b217beffa53f.

kairos-cli/bin/crypto/private_key.rs
koxu1996: Thanks, improved in https://github.com/cspr-rad/kairos/pull/17/commits/b1a78b47dfce7f834e6d5f05e824b217beffa53f.

kairos-cli/bin/crypto/signer.rs
koxu1996: Done in https://github.com/cspr-rad/kairos/pull/17/commits/8b78660571180b15f5319049925eb6431b0ced1d.

kairos-cli/bin/crypto/public_key.rs
koxu1996: Removed in https://github.com/cspr-rad/kairos/pull/17/commits/72d333c22d8de19c00ea4debfa6d9a3a808fb893.

kairos-cli/bin/error.rs
koxu1996: Good catch! Outdated TODO removed in https://github.com/cspr-rad/kairos/pull/17/commits/67be402a1cdc9dd9f3eb60875e06d1a8d00322d8.

kairos-cli/bin/common/args.rs
koxu1996: I tried that... but it appeared that `Clone` must be implemented for the argument, which is not present in Casper types:

```
error[E0277]: the trait bound `casper_types::SecretKey: std::clone::Clone` is not satisfied
 --> kairos-cli/bin/crypto/private_key.rs:6:29
  |
5 | #[derive(Clone)]
  |          ----- in this derive macro expansion
6 | pub struct CasperPrivateKey(pub casper_types::SecretKey);
  |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::clone::Clone` is not implemented for `casper_types::SecretKey`

**Comment by koxu1996** [comment]
@marijanp

> Great, I think we are almost there. One last remark. I've put the `main.rs` into build to create a separation between library and binary. If you could put all the modules except `main.rs` into a `src` directory that would be perfect.

Okay! Files moved in https://github.com/cspr-rad/kairos/pull/17/commits/152858a81f343cec0a82d4dda74182a88ae5e571.

File: kairos-cli/bin/commands/mod.rs (hunk 19 above)
Comment (koxu1996) [reply kairos-cli/bin/commands/mod.rs]: In case we put CLI 'spec' in one file, then it will become like:

```rs
#[derive(Subcommand)]
pub enum Command {
    #[command(about = "Deposits funds into your account")]
    Deposit {
        #[clap(flatten)]
        amount: AmountArg,
        #[clap(flatten)]
        private_key_path: PrivateKeyPathArg,
    },
    // ...
```

Looks great... but how we can get `DepositArgs` type for deposit handler `pub fn run(args: DepositArgs) -> Result<String, CliError>`?

Maybe some handler-derive macro could work, but I could not find a way to make it work :roll_eyes:.

I thought about different files structure, but merging all those commands into one file does not make sense to me. I would leave it as it is.

kairos-cli/src/crypto/private_key.rs
```diff
@@ -0,0 +1,13 @@
...
+
+use super::error::CryptoError;
+
+pub struct CasperPrivateKey(pub casper_types::SecretKey);
+
+impl CasperPrivateKey {
+    pub fn from_file<P: AsRef<Path>>(file_path: P) -> Result<Self, CryptoError> {
+        casper_types::SecretKey::from_file(file_path)
+            .map(Self)
+            .map_err(|error| error.into())
```
Avi-D-coder, line 11: Return `ErrorExt` not the more general `CryptoError`.
We want our errors to be as precise as possible within reason, such that if you match on it you don't have to handle impossible error cases.

Ideally a function's error enum should only contain variants that the function may construct.
This ideal is not always the reasonable, a frequent compromise is to write an error per module.
`ErrorExt` has variant's `casper_types::SecretKey::from_file` never constructs, but I would not suggest spending time fixing it.

kairos-cli/src/crypto/signer.rs
```diff
@@ -0,0 +1,38 @@
...
+use casper_types::{crypto, PublicKey};
+
+pub struct CasperSigner {
+    secret_key: CasperPrivateKey,
+    pub public_key: CasperPublicKey,
+}
+
+#[allow(unused)]
+impl CasperSigner {
+    pub fn from_file<P: AsRef<Path>>(file_path: P) -> Result<Self, CryptoError> {
```
Avi-D-coder, line 16: Same as above

Avi-D-coder (changes requested):

kairos-cli/src/crypto/private_key.rs
koxu1996: My intention was to isolate Casper related types in `crypto` module, which introduced intermediary `CryptoError`. Returning `ErrorExt` directly would make all 3 command handlers more complex, because error has to be upcasted manually:

```rs
let _signer = CasperSigner::from_file(args.private_key_path.field)
        .map_err(|error| CliError::CryptoError { error: error.into() })?;
```

We could avoid it by merging `CryptoError` with `Error`, but not sure if this is the thing you want.

Avi-D-coder: Yeah you would have to add `.map_err(CryptoError::from)` to each handler.

> If I update crypto related function to return ErrorExt directly, then I am polluting whole project with Casper specific type
If the goal is to encapsulate `casper_types` like `ErrorEx` to guard callers against future upstream changes then `ErrorExt` cannot appear in CryptoError at all. It may be nested, but anything in a enum is public.

The point of using a static error type (the thiserror style) is to be able to match on the errors and handle exactly what went wrong. An overly general error return does not tell the caller what might go wrong. This leaves the caller to look through the code to find out what each function might return; like I did with `ErrorExt`.

We may not want precise errors, maybe we're too early, or we don't expect anyone to use them. 
Given this is CLI app code not library code we could choose to use `anyhow` or another dynamic error type.
If we are going to have general errors I'd suggest that for future work.

For now I'd be okay with leaving the return as `CryptoError` and changing it when we need it.

kairos-cli/src/crypto/signer.rs
koxu1996: Fixed in https://github.com/cspr-rad/kairos/commit/55560461a0781fef8734bdd43b0118fb740ea9ec.

Avi-D-coder (approved):

Avi-D-coder:
LGTM

marijanp (approved):
LGTM

kairos-cli/src/crypto/signer.rs
```diff
@@ -0,0 +1,38 @@
...
+    pub public_key: CasperPublicKey,
+}
+
+#[allow(unused)]
+impl CasperSigner {
+    pub fn from_file<P: AsRef<Path>>(file_path: P) -> Result<Self, ErrorExt> {
+        let secret_key = CasperPrivateKey::from_file(file_path)?;
+
+        // Derive the public key.
+        let public_key = CasperPublicKey(PublicKey::from(&secret_key.0));
```
quinn-dougherty, line 20: I would have considered separating file IO from retrieval of public key. I'd like granular control over error whether `from_file` fails because the file isn't there or because it's ill-formatted (which presumably would make `PublicKey::from` fail), and separating it into more smaller functions would be one way to do this.

quinn-dougherty (approved):
LGTM

kairos-cli/src/crypto/signer.rs
koxu1996: 1. `PublicKey::from` cannot fail.

2. `CasperSigner::from_file` already provides granular errors, for example:

- non existing file :arrow_right: `ErrorExt::SecretKeyLoad(ReadFileError { ... })`
- invalid PEM content :arrow_right: `ErrorExt::FromPem("malformedframing")`

kairos-cli/src/commands/deposit.rs
```diff
@@ -0,0 +1,26 @@
+use crate::common::args::{AmountArg, PrivateKeyPathArg};
+use crate::crypto::error::CryptoError;
+use crate::crypto::signer::CasperSigner;
```
xcthulhu, line 3: Why did we make a custom signer?

kairos-cli/src/crypto/public_key.rs
```diff
@@ -0,0 +1,24 @@
+use crate::crypto::error::CryptoError;
+use casper_types::bytesrepr::{FromBytes, ToBytes};
+
+#[derive(Clone)]
+pub struct CasperPublicKey(pub casper_types::PublicKey);
+
+impl CasperPublicKey {
+    pub fn from_bytes(bytes: &[u8]) -> Result<Self, CryptoError> {
```
xcthulhu, line 8: Don’t make this part of the inherent implementation, just impl the FromBytes trait if you want to do this

kairos-cli/src/crypto/signer.rs
```diff
@@ -0,0 +1,38 @@
...
+
+use super::private_key::CasperPrivateKey;
+use super::public_key::CasperPublicKey;
+use crate::crypto::error::CryptoError;
+use casper_types::bytesrepr::ToBytes;
+use casper_types::{crypto, ErrorExt, PublicKey};
+
+pub struct CasperSigner {
+    secret_key: CasperPrivateKey,
+    pub public_key: CasperPublicKey,
```
xcthulhu, line 11: Why do we need to have a public key here?

kairos-cli/src/crypto/public_key.rs
koxu1996: `FromBytes` is a trait existing in `casper_types`, and my intention was to make Kairos as much independent as possible.

kairos-cli/src/crypto/public_key.rs
```diff
@@ -0,0 +1,24 @@
...
+        let (public_key, _remainder) =
+            casper_types::PublicKey::from_bytes(bytes).map_err(|_e| {
+                CryptoError::Serialization {
+                    context: "public key",
+                }
+            })?;
+        Ok(Self(public_key))
+    }
+
+    #[allow(unused)]
```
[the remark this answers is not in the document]
koxu1996: Same as with `FromBytes` - this would make Kairos more entangled with Casper's code :roll_eyes:. I could reimplement `ToBytes` trait, but it does not seem useful at this point.

kairos-cli/src/commands/deposit.rs
koxu1996: I want to isolate Casper code, to make it easily replaceable in the future. I imagine we might have `Signer` trait and blockchain specific implementations for it.

kairos-cli/src/crypto/signer.rs
koxu1996: Public key is necessary for [signing data with Casper signer](https://docs.rs/casper-types/4.0.1/casper_types/crypto/fn.sign.html):

```rs
pub fn sign<T: AsRef<[u8]>>(
    message: T,
    secret_key: &SecretKey,
    public_key: &PublicKey
) -> Signature
```

Since we have user's private key, we can derive public key - we could do that every time `sign_message()` is called, but I thought storing it might come handy soon.

kairos-cli/tests/cli_tests.rs
```diff
@@ -0,0 +1,113 @@
...
+    cmd.assert().success().stdout("ok\n");
+}
+
+#[test]
+fn deposit_invalid_amount() {
+    let secret_key_path = fixture_path("ed25519/secret_key.pem");
```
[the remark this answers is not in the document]
koxu1996: @xcthulhu I was not considering ZK yet. Most Casper users are using `ed25519` keys, and I wanted to make it easy for them to use Kairos without additional steps. If we want to use different algorithm, then we have to think about mapping it to Casper keys, but I think this requires more discussion.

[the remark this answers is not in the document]
koxu1996: Notably, Secp256k1 keys - mentioned in #7 - are also supported by current implementation :wink:.

kairos-cli/src/crypto/signer.rs
xcthulhu: Ahh... don't use that signer, it has an unused public key for some reason.

Use this instead:

```rust
pub fn sign<T: AsRef<[u8]>>(secret_key: &SecretKey, message: T) -> Signature {
    match secret_key {
        SecretKey::System => Signature::System,
        SecretKey::Ed25519(secret_key) => {
            let signature = secret_key.sign(message.as_ref());
            Signature::Ed25519(signature)
        }
        SecretKey::Secp256k1(secret_key) => {
            let signature = secret_key
                .try_sign(message.as_ref())
                .expect("should create signature");
            Signature::Secp256k1(signature)
        }
        _ => panic!("SecretKey is marked as non-exhaustive, but this should never happen"),
    }
}
```

kairos-cli/src/crypto/public_key.rs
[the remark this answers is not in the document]
xcthulhu: Hmm... but if you aren't using it, do we need it for anything?

kairos-cli/tests/cli_tests.rs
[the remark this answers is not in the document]
xcthulhu: > @xcthulhu I was not considering ZK yet.

I'd really recommend we do some feasibility research into this first before proceeding much further.

kairos-cli/src/commands/transfer.rs
```diff
@@ -0,0 +1,31 @@
...
+    #[arg(long, short, value_name = "PUBLIC_KEY", value_parser = parse_hex_string)]
+    recipient: ::std::vec::Vec<u8>, // Absolute path is required here - see https://github.com/clap-rs/clap/issues/4626#issue-1528622454.
+    #[clap(flatten)]
+    amount: AmountArg,
+    #[clap(flatten)]
+    private_key_path: PrivateKeyPathArg,
+}
+
+pub fn run(args: Args) -> Result<String, CliError> {
+    let _recipient = CasperPublicKey::from_bytes(args.recipient.as_ref())?;
```
xcthulhu, line 21: So... do we have a test that public keys on https://cspr.live/ parse using this?

kairos-cli/src/crypto/signer.rs
```diff
@@ -0,0 +1,38 @@
...
+}
+
+#[allow(unused)]
```
xcthulhu, line 14: Isn't this used though?

koxu1996: Not used, because we have no signature verification yet:

https://github.com/cspr-rad/kairos/blob/2ef83df22bbdfa06f2d49b8fed309c0274ba907d/kairos-server/src/routes/transfer.rs#L34

This is about to come soon - once we establish transaction format (I am currently exploring protobuffs and ASN.1+DER).

koxu1996:
Changes rebased with the latest `main`.

xcthulhu (dismissed):

kairos-cli/src/commands/transfer.rs
koxu1996: Tests added in https://github.com/cspr-rad/kairos/pull/17/commits/3775a71005d2f99a8e9dbe9d158ea16c052abab2.

koxu1996:
Per @xcthulhu request, I added additional tests for parsing public keys from CSPR.live. All four approvals were automatically dismissed though :fearful:.

@asladeofgreen PTAL :pray:

marijanp (approved):

kairos-cli/src/crypto/public_key.rs
```diff
@@ -0,0 +1,68 @@
...
+
+#[derive(Clone)]
+pub struct CasperPublicKey(pub casper_types::PublicKey);
+
+impl CasperPublicKey {
+    pub fn from_bytes(bytes: &[u8]) -> Result<Self, CryptoError> {
+        let (public_key, _remainder) =
+            casper_types::PublicKey::from_bytes(bytes).map_err(|_e| {
+                CryptoError::Serialization {
+                    context: "public key",
```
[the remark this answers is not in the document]
koxu1996: Improved in https://github.com/cspr-rad/kairos/pull/17/commits/2a3771802399a457f53bd5851a5b39edccdd9a86.

kairos-cli/src/crypto/public_key.rs
```diff
@@ -0,0 +1,68 @@
...
+                }
+            })?;
+        Ok(Self(public_key))
+    }
+
+    #[allow(unused)]
+    fn to_bytes(&self) -> Result<Vec<u8>, CryptoError> {
+        self.0.to_bytes().map_err(|_e| CryptoError::Serialization {
+            context: "public key",
```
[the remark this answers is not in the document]
koxu1996: Improved in https://github.com/cspr-rad/kairos/pull/17/commits/2a3771802399a457f53bd5851a5b39edccdd9a86.

marijanp (approved):

Avi-D-coder (approved):
lgtm

asladeofgreen (approved):
I have caveats, notably with reinventing wheels in respect of the cryptography, however I do understand the wish to reduce dependency upon casper libs as much as possible.  Wouldn't a dedicated low-level kairos-crypto crate be of use here ?

So am approving for now but imho would like to revisit some of this code as the solution evolves.

koxu1996:
> I have caveats, notably with reinventing wheels in respect of the cryptography, however I do understand the wish to reduce dependency upon casper libs as much as possible. Wouldn't a dedicated low-level kairos-crypto crate be of use here ?

@asladeofgreen Yes, cryptography will be moved out of CLI package into `kairos-crypto` in https://github.com/cspr-rad/kairos/pull/35. It provides `pub trait CryptoSigner`, with Casper implementation under feature flag.

PR #27, opened 2024-02-21
ci: configure hercules-ci to only build x86_64-linux

marijanp:


quinn-dougherty (approved):
Not e2e verified yet cuz box keeps getting OOM kills, but have confirmed that the job dispatch is now omitting the darwin jobs. 

EDIT: all went through

Avi-D-coder (approved):

PR #26, opened 2024-02-20
add kairos deposit contract and contract-cli

jonas089:
Kairos deposit contract and kairos-contract-cli

For instructions on how to build and unit-test the contract, navigate to

`kairos-deposit-contract/readme.md`

The kairos-contract-cli can be used to interact with the contract once it has been deployed to cctl (see cspr-rad/kairos-cctl-env).

PR #28, opened 2024-02-23
Package deposit contract and test with nix

marijanp:


marijanp:
Have to close this because the original deposit contract PR #26  was closed

PR #29, opened 2024-02-29
add deposit contract crates to workspace

jonas089:
+ There is a dedicated crate for contract tests
+ The deposit smart contract fetches node and casper-client dependencies from cspr-rad/casper-node & cspr-rad/casper-client

The compile-contracts.sh script can be used to build all session code that is used for testing, as well as the smart contract itself.

The PATH_TO_WASM_BINARIES environment variable must be set before running the tests. An example can be found in the deposit contract's readme.md

The deployment target is the latest feature branch of the Casper node 2.0 (28th of Feb 2024) and a seemingly stable release of casper-client. Contract deployment was tested/ can be tested using the kairos-cctl-env.


kairos-deposit-contract/compile-contracts.sh
```diff
@@ -0,0 +1,16 @@
+rustup target add wasm32-unknown-unknown
+
+cargo build -p contract --release --target wasm32-unknown-unknown
+wasm-opt --strip-debug --signext-lowering ./target/wasm32-unknown-unknown/release/deposit-contract.wasm -o ./target/wasm32-unknown-unknown/release/deposit-contract-optimized.wasm
```
koxu1996, line 4: This introduces `wasm-opt` dependency that has to be manually installed :unamused:.

I believe we can remove it from this bash script, and include [`wasm_opt` crate](https://docs.rs/wasm-opt/0.116.0/wasm_opt/) in the tests, that will be used for creating optimized file side-by-side (unfortunately the library does not allow working directly on memory).

jonas089: Installing wasm-opt system wide is a viable option imo. Anyone working with casper contracts should have it available anyways. Makefiles in other contracts rely on it as well and it is as long as it works I don't see a need to change anything about it.

kairos-deposit-contract/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[workspace]
+resolver = "2"
+
+members = [
+  "malicious-reader",
+  "malicious-session",
+  "withdrawal-session",
+  "deposit-session",
```
koxu1996, line 8: What is the use of those session codes?

jonas089:
> > The deposit smart contract fetches node and casper-client dependencies from cspr-rad/casper-node & cspr-rad/casper-client
> 
> @jonas089 What is exactly changed in the custom version of `casper-node` and `casper-client`? Btw. client does not seem to be used.

Casper-client is used in my contract-cli / library that I excluded from this PR as per your request.
The casper-node and casper-client are forked so that we have a version-locked development branch for Kairos.
I did some quality of life improvements that you can take a look at, but these details are not important for Kairos development.

kairos-deposit-contract/Cargo.toml
jonas089: "Malicious": are used for testing malicious behavior when some unauthorized tries to withdraw from the contract purse

"Deposit-session and Withdrawal-session": Session code to deposit and withdraw from the contract.

koxu1996:
> > > The deposit smart contract fetches node and casper-client dependencies from cspr-rad/casper-node & cspr-rad/casper-client
> > 
> > 
> > @jonas089 What is exactly changed in the custom version of `casper-node` and `casper-client`? Btw. client does not seem to be used.
> 
> Casper-client is used in my contract-cli / library that I excluded from this PR as per your request. The casper-node and casper-client are forked so that we have a version-locked development branch for Kairos. I did some quality of life improvements that you can take a look at, but these details are not important for Kairos development.

I would like to know what exactly are the improvements, because if the only change is single WASM path, then I think it's not worth maintaining the custom version of node (+ client).

kairos-deposit-contract/Cargo.toml
jonas089: Since 2.0 is not done yet we have no option to spend funds from the caller purse without using session code.

jonas089: When session code gets deprecated we just replace the session code deploy with a contract call, not a big deal.

jonas089: Ultimately it will all depend on future build targets. For now we have the build target that is locked in the kairos-* branches.

koxu1996:
Pardon me, but I see that using custom node is useless now (if the only change is WASM path). If there is no reason to maintain custom fork, then I am heavily against relying on it. It would be helpful if you could describe those changes, so everyone could understand why we need them.

I understand your frustration with the amount of work to make changes nice and clean, but one of the project pillars was to focus on clean and production ready code.

jonas089:
> Pardon me, but I see that using custom node is useless now (if the only change is WASM path). If there is no reason to maintain custom fork, then I am heavily against relying on it. It would be helpful if you could describe those changes, so everyone could understand why we need them.
> 
> I understand your frustration with the amount of work to make changes nice and clean, but one of the project pillars was to focus on clean and production ready code.

We clearly don't have the same definition of clean code on this matter. We are in a development environment with no definite build target, therefore I don't see a need to be this precise when it comes to the build target for prototyping. You didn't want me to use commit hashs in the Cargo.toml and suggested to version-lock by creating a branch under cspr-rad. This is what I did and yes it cost me a significant amount of time to achieve the same result. I changed the wasm output path in the test engine because on the main repo it is derived from a workspace constant that is set by cargo automatically. This is only useful when the contract is a seperate project and not when the tests are included in a higher-level workspace as is the case with kairos. The change is non-breaking and the test engine has no impact on the production wasm code execution. We are definitely better off with a version locked target since breaking changes can occur on the feature branch and pointing to the main repo will be a huge pain. We want clean code, but we also want a working prototype. We need to make some compromises in the beginning as we can spend a lot of time refining when the fundamentals are in place.

koxu1996:
@jonas089 I didn't request any changes yet. Before I can submit review, I would like to understand your changes - especially what needs to be adjusted in `casper-node` to make it work for Kairos, because I believe we can simply avoid it. Please help me understanding it, then we will get it merged sooner.

Regarding commit-locking, I think you should see [Specifying dependencies from git repositories](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories). Changes happening on the branch will not break our code, as they are locked with `Cargo.lock` :wink::

> Once a git dependency has been added, Cargo will lock that dependency to the latest commit at the time. New commits will not be pulled down automatically once the lock is in place. However, they can be pulled down manually with cargo update.

jonas089:
Someone could override Cargo.lock or run cargo update, I think a seprate branch is more secure.

jonas089:
> @jonas089 I didn't request any changes yet. Before I can submit review, I would like to understand your changes - especially what needs to be adjusted in `casper-node` to make it work for Kairos, because I believe we can simply avoid it. Please help me understanding it, then we will get it merged sooner.
> 
> Regarding commit-locking, I think you should see [Specifying dependencies from git repositories](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories). Changes happening on the branch will not break our code, as they are locked with `Cargo.lock` :wink::
> 
> > Once a git dependency has been added, Cargo will lock that dependency to the latest commit at the time. New commits will not be pulled down automatically once the lock is in place. However, they can be pulled down manually with cargo update.

As far as the Casper-Node goes the only change that I had to make is the one in the execution-engine-tests regarding the wasm path (and the fact that I would rather lock the version by branch than rely on Cargo.lock).

koxu1996 (changes requested):
> As far as the Casper-Node goes the only change that I had to make is the one in the execution-engine-tests regarding the wasm path (and the fact that I would rather lock the version by branch than rely on Cargo.lock).

Okay, I am glad we confirmed it's the only change. We can untangle this then :smiley:, please follow.  I guess you customized `casper-node` in order to make this code works in the `test_fixture.rs`:

```rs
    pub fn install(&mut self, admin: AccountHash) {
        ...
        let install_contract_request =
            ExecuteRequestBuilder::standard(admin, "deposit-contract-optimized.wasm", session_args)
                .build()
```

If you check [how WASM paths are resolved underneath](https://docs.rs/casper-engine-test-support/7.0.1/src/casper_engine_test_support/utils.rs.html#106-134), you will see that it:

1. Firstly, if path is relative, it goes through the predefined paths (including the one you customized) and check if file exists.
2. Secondly, **it checks if file exists at given relative path (to the current directly) or absolute path**.

:bulb: So there is no point in customizing predefined paths, just provide valid relative/absolute path directly in tests.

koxu1996:
> Someone could override Cargo.lock or run cargo update, I think a seprate branch is more secure.

Forking repository just for locking changes does not make sense. Just use Cargo's lock `branch`; or even `rev` if you are really worried that someone who is updating dependencies does not expect to update dependencies.

kairos-deposit-contract/malicious-session/src/main.rs
```diff
@@ -0,0 +1,15 @@
+#![no_std]
+#![no_main]
+use casper_contract::contract_api::{account, runtime, system};
+use casper_types::{runtime_args, AddressableEntityHash, URef, U512};
+
```
marijanp, line 5: Can we add a comment here on what a `malicious-session` does/ is supposed to do?

kairos-deposit-contract/malicious-session/Cargo.toml
marijanp, line 1: missing newline at the end of file. I will stop pointing it out, the github diffs point it out

kairos-deposit-contract/malicious-reader/src/main.rs
```diff
@@ -0,0 +1,12 @@
+#![no_std]
+#![no_main]
+use casper_contract::contract_api::{account, runtime, system};
+use casper_types::{URef, U512};
+
```
marijanp, line 5: Please add a comment on what the malicious reader is supposed to do.

kairos-deposit-contract/deposit-session/src/main.rs
```diff
@@ -0,0 +1,24 @@
...
+#[no_mangle]
+pub extern "C" fn call() {
+    let contract_hash: AddressableEntityHash = runtime::get_named_arg("deposit_contract");
+    let amount: U512 = runtime::get_named_arg("amount");
+    let source: URef = account::get_main_purse();
+    // create a temporary purse that can be passed to the deposit contract
+    let temp_purse: URef = system::create_purse();
+    // fund the temporary purse
+    system::transfer_from_purse_to_purse(source, temp_purse, amount, None).unwrap();
+    // call the deposit endpoint
```
marijanp, line 15: the code literally says `call_contract(..."deposit")`

kairos-deposit-contract/deposit-session/src/main.rs
```diff
@@ -0,0 +1,24 @@
...
+#![no_main]
+use casper_contract::contract_api::{account, runtime, system};
+use casper_types::{runtime_args, AddressableEntityHash, URef, U512};
+
+#[no_mangle]
+pub extern "C" fn call() {
+    let contract_hash: AddressableEntityHash = runtime::get_named_arg("deposit_contract");
+    let amount: U512 = runtime::get_named_arg("amount");
+    let source: URef = account::get_main_purse();
+    // create a temporary purse that can be passed to the deposit contract
```
marijanp, line 11: It would be good if you can explain what purpose the temporary purse serves. Do we need an intermediate purse to deposit to our contract's purse? is it then `owner_purse` -> `tem_purse` -> `contract_purse`

marijanp (changes requested):

kairos-deposit-contract/compile-contracts.sh
koxu1996: >  Anyone working with casper contracts should have it available anyways.

@jonas089 Wrong. **You** have it installed, because I showed you this can be used for [fixing opcode 192 issue](https://github.com/casper-network/casper-node/issues/4367#issuecomment-1798438389) and using most recent Rust complier. By default Casper contracts rely on `wasm-strip` (binary from `wabt`), so I am still insisting on using crate instead of unknown system binary.

jonas089: Since you know exactly what to do here and I don't, would it be possible for you to submit a PR over this issue? Otherwise I can look into it later today.

kairos-deposit-contract/compile-contracts.sh
```diff
@@ -0,0 +1,16 @@
+rustup target add wasm32-unknown-unknown
```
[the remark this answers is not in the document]
jonas089: bash

kairos-deposit-contract-tests/src/test_fixture/utils.rs
```diff
@@ -0,0 +1,46 @@
...
+    DEFAULT_ACCOUNT_INITIAL_BALANCE,
+};
+use casper_types::{
+    account::AccountHash,
+    runtime_args,
+    system::{handle_payment::ARG_TARGET, mint::ARG_ID},
+    PublicKey, SecretKey,
+};
+
+// Creates a dummy account and transfer funds to it
```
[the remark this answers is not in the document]
jonas089: Since this is the tests crate, the case where the account_string is None doesn't need to be handled. We just need a set of n dummy accounts to work with, there are no edgecases.

kairos-deposit-contract/malicious-session/src/main.rs
jonas089: I will add a comment explaining the session codes on top of each file.

kairos-deposit-contract/malicious-session/Cargo.toml
jonas089: Is there a quick and easy fix or will I have to do it manually? I ran nix fmt but apparently that doesn't help here?

kairos-deposit-contract/malicious-reader/src/main.rs
jonas089: I will resolve this since I will add comments to all session code as mentioned above. I will keep the other conversation open.

kairos-deposit-contract/deposit-session/src/main.rs
jonas089: The entry point name is static, I can introduce a constant if you prefer that. Something like DEPOSIT_EP_NAME = "deposit".

kairos-deposit-contract/compile-contracts.sh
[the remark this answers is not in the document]
marijanp: When someone uses another default shell than bash? Someone could be running fish per default for example.

kairos-deposit-contract-tests/src/test_fixture/utils.rs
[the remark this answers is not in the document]
marijanp: That is exactly what I mean. I'm pointing out that these non-obvious conditions need to be documented, such that any developer using or extending this function understands the intention. I didn't know this, and I still dont understand what the `None` case does in case the account string is None.
If you say that the None case is impossible, then it would be better to design the interface that way or maybe use `expect`

kairos-deposit-contract/malicious-session/Cargo.toml
marijanp: I've added an editorconfig to this repository. Your editor should be able to pick that up and add these newlines.

kairos-deposit-contract/deposit-session/src/main.rs
marijanp: I haven't been clear enough with my comment: I mean documentation should be used to point out non-obvious things. This comment is repeating the code.

kairos-deposit-contract/deposit-session/src/main.rs
marijanp: Please document this here such that someone revisiting this code is aware

kairos-deposit-contract/deposit-session/src/main.rs
jonas089: Wouldn't it be easier to suggest a change directly than open a conversation in cases where you only want a single line to be removed or edited?

kairos-deposit-contract-tests/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "kairos-deposit-contract-tests"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
```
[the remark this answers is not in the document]
jonas089: Done.

kairos-deposit-contract/readme.md
```diff
@@ -0,0 +1,112 @@
...
+
+1. Call the session code to transfer N tokens from main_purse to the freshly created temp_purse.
+
+2. The session code will then call the deposit entry point of the deposit contract and pass the temp_purse as a reference.
+
+3. All funds from the temp_purse will be transferred to the contract_purse that is managed by the ADMIN.
+
+
+## Building and testing
+Currently the paths in build-macos-darwin.sh are configured for Macos, nixifying this makes sense. Paths must be discoverable by utils.rs in the test engine.
```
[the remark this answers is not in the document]
jonas089: Paths to the compiled wasm binaries. This is why we set the environment variable PATH_TO_WASM_BINARIES=...

kairos-deposit-contract/readme.md
```diff
@@ -0,0 +1,112 @@
...
+
+For now, on Macos:
+
+`./build-macos-darwin-feat-2.0.sh`
```
marijanp, line 93: This file is missing

PR #30, opened 2024-03-04
Deterministic payload format: ASN.1 schema with DER encoding

koxu1996:
# Deterministic payload format

This PR defines fully deterministic payload format for Kairos - based on **ASN.1** :heart:  - that can replace currently used JSON. All relevant code for server integration is provided, although integration will be done in the subsequent PR.

### Motivation

Since we want to sign payloads with private key, we have to make sure format is fully deterministic. Otherwise there might be issue with signature verification across different languages/platforms. This is also future requirement for building ledger, as we need mechanism for _transaction hashing_ :magic_wand:.

### Why not Canonical JSON, or Protocol Buffers?

They were not designed with determinism in mind, and it makes them non-applicable to blockchain domain. Ethereum uses [_RLP_](https://ethereum.org/en/developers/docs/data-structures-and-encoding/rlp/), Casper uses _CLTypes_, ... and I think Kairos could use serialization based on **ASN.1 with DER encoding**.

### What is ASN.1?

To get a quick grasp, I would say it is like Protocol Buffers that were created 40 years ago for the use in telecommunication. They say ASN.1 works so well, that it is used everywhere, but no one even knows about it :wink:.

I highly recommend reading the introduction of [Warm Welcome to ASN.1 and DER](https://letsencrypt.org/docs/a-warm-welcome-to-asn1-and-der/), as it explains the basics.

## ASN.1 schema

Schema was prepared according to the standards - you can compile it with [online playground](https://asn1.io/asn1playground/default.aspx)  and verify there is no warning and no error. It is also forward compatible - used `---` for `CHOICE` and `SEQUENCE` - so more fields can be added in the future.

The following schema is available at `./kairos-tx/schema.asn`:

```
--<Kairos ANS.1 Schema>--

TxSchema DEFINITIONS AUTOMATIC TAGS ::= BEGIN

  -- Basic types.
  PublicKey ::= OCTET STRING
  Amount ::= INTEGER (0..18446744073709551615)
  Nonce ::= INTEGER (0..18446744073709551615)

  -- Transaction payload for signing.
  SigningPayload ::= SEQUENCE {
    nonce Nonce,
    body TransactionBody,
    ...
  }

  -- Generic transaction body.
  TransactionBody ::= CHOICE {
    deposit [0] Deposit,
    transfer [1] Transfer,
    withdrawal [2] Withdrawal,
    ...
  }

  -- Detailed definitions of each transaction type.
  Deposit ::= SEQUENCE {
    amount Amount,
    ...
  }

  Transfer ::= SEQUENCE {
    recipient PublicKey,
    amount Amount,
    ...
  }

  Withdrawal ::= SEQUENCE {
    amount Amount,
    ...
  }

END
```

**Note:** Amount and nonce are constrained to be in u64 range.

## Example payloads

Payloads are encoded with DER, which is binary format. For simplicity I provide PEM files (base64-armored DER + label), that can be easily analyzed online - https://lapo.it/asn1js/#MAkCAQGiBAICA-g: 

image

### 1. Deposit

- nonce: `1`
- amount: `1000`

**PEM formatted data:**

```
-----BEGIN KAIROS PAYLOAD-----
MAkCAQGgBAICA+g=
-----END KAIROS PAYLOAD-----
```

Raw data size: 11 bytes.

### 2. Transfer

- nonce: `1`
- recipient: `0B0B0B0B0B0B0B0B0B0B0B0B0B0B0B0B0B0B0B0B0B0B0B0B0B0B0B0B0B0B0B0B`
- amount: `1000`

**PEM formatted data:**

```
-----BEGIN KAIROS PAYLOAD-----
MCsCAQGhJgQgCwsLCwsLCwsLCwsLCwsLCwsLCwsLCwsLCwsLCwsLCwsCAgPo
-----END KAIROS PAYLOAD-----
```

Raw data size: 45 bytes.

### 3. Withdrawal

- nonce: `1`
- amount: `1000`

**PEM formatted data:**

```
-----BEGIN KAIROS PAYLOAD-----
MAkCAQGiBAICA+g=
-----END KAIROS PAYLOAD-----
```

Raw data size: 11 bytes.


Avi-D-coder (approved):
LGTM

kairos-tx/src/crypto.rs
marijanp, line 1: Where is this crypto module used?

kairos-tx/src/parsers.rs
marijanp, line 1: These are not really parsers and I would prefer them to be closer to the actual type definition.

marijanp (commented):
Great PR description, it was very easy to follow your intentions. Looks good to me, I only two small comments.

kairos-tx/src/crypto.rs
koxu1996: Good catch! I have used it for testing integration with `kairos-server`, but this should not be included in this PR. Removed in https://github.com/cspr-rad/kairos/pull/30/commits/99dc80003b523d846dca6d4ec6de558f4af6c5de.

kairos-tx/src/parsers.rs
koxu1996: Okay, moved in https://github.com/cspr-rad/kairos/pull/30/commits/a7698115069dfaadd7f0518a50bd2ab69424fe6d.

marijanp (approved):

koxu1996:
PTAL @asladeofgreen or @xcthulhu :pray:.

Avi-D-coder (approved):
LGTM

koxu1996:
Just for clarity: we relaxed the rules of merging and 2 approvals from anyone from the team are enough now.

Issue #31, opened 2024-03-04
String (utf-8) used for public key and signature


koxu1996:
Both _public key_ and _signature_ types are defined as `String`:

https://github.com/cspr-rad/kairos/blob/2ef83df22bbdfa06f2d49b8fed309c0274ba907d/kairos-server/src/lib.rs#L12-L13

This is clearly wrong, as those are arbitrary bytes, not necessarily valid UTF-8 data.

PR #32, opened 2024-03-04
Replace `PublicKey` and `Signature` with plain `Vec<u8>`

koxu1996:
Fix for #31 - we want to rely on bytes, not strings. Additionally, using type aliases at this point has no value.

### DIFF
--- a/kairos-server/src/lib.rs
+++ b/kairos-server/src/lib.rs
@@ -9,8 +9,8 @@ use state::LockedBatchState;
 
 pub use errors::AppErr;
 
-type PublicKey = String;
-type Signature = String;
+type PublicKey = Vec<u8>;
+type Signature = Vec<u8>;
 
 pub fn app_router(state: LockedBatchState) -> Router {
     Router::new()
--- a/kairos-server/src/routes/deposit.rs
+++ b/kairos-server/src/routes/deposit.rs
@@ -42,7 +42,7 @@ pub async fn deposit_handler(
     *balance = updated_balance;
 
     tracing::info!(
-        "Updated account public_key={} balance={}",
+        "Updated account public_key={:?} balance={}",
         public_key,
         updated_balance
     );
--- a/kairos-server/src/routes/withdraw.rs
+++ b/kairos-server/src/routes/withdraw.rs
@@ -63,7 +63,7 @@ pub async fn withdraw_handler(
     }
 
     tracing::info!(
-        "Updated account public_key={} balance={}",
+        "Updated account public_key={:?} balance={}",
         withdrawal.public_key,
         updated_balance
     );


Avi-D-coder (approved):
LGTM

quinn-dougherty (approved):
straightforward, lgtm --- I look forward to reading more your arguments philosophically for value prop of not type aliasing!

marijanp:
Well I don't agree that the type aliasing or even actually better newtyping does not add any value. It adds semantics and protects the user of the . Now I send an amount of tokens from `Vec<u8>` to `Vec<u8>` which could be anything that is representable as bytes. Obscurity is a root cause of complexity and obscurity occurs when important information is not obvious.

koxu1996:
@marijanp  I agree that newtypes provide beneficial semantic value, but... in this case I **just felt** it would be better to remove them for now:

> using type aliases *at this point* has no value

If you consider this without more context, it might sounds strange at first, but I will try to explain what happened in my head.

_Btw. I think it is harder to justify why not to use some great thing, rather than justifying to use it._

:cactus: Reason to remove `type Signature` is simple. It is used only in the one place, for handling transfer's request body:

https://github.com/cspr-rad/kairos/blob/2ef83df22bbdfa06f2d49b8fed309c0274ba907d/kairos-server/src/routes/transfer.rs#L14-L25

:arrow_right: There is no signing mechanism yet, so I would not worry about losing information or increasing complexity here. However, with https://github.com/cspr-rad/kairos/pull/30 (already approved by Avi) we will have payload structure that will contain signature embedded **inside**, and no one will be interacting with signature directly - always via payload structure.

:cactus: Regarding `type PublicKey`, it is currently used for two things:

1) Handling transfer's request body.
    
    :arrow_right:  Same story as with the signature - it will become obsolete, once we integrate new payload structure.
    
2) Storing balances:
    
    https://github.com/cspr-rad/kairos/blob/2ef83df22bbdfa06f2d49b8fed309c0274ba907d/kairos-server/src/state.rs#L16-L17
    
    :arrow_right: If we consider this is the only place where `PublicKey` newtype is used, then I am not happy with seeing it inside `kairos-server` crate. I believe it could be placed in `kairos-types` or `kairos-crypto`, but at this point using it seems to provide no value.

cc @quinn-dougherty

jonas089:
> @marijanp I agree that newtypes provide beneficial semantic value, but... in this case I **just felt** it would be better to remove them for now:
> 
> > using type aliases _at this point_ has no value
> 
> If you consider this without more context, it might sounds strange at first, but I will try to explain what happened in my head.
> 
> _Btw. I think it is harder to justify why not to use some great thing, rather than justifying to use it._
> 
> 🌵 Reason to remove `type Signature` is simple. It is used only in the one place, for handling transfer's request body:
> 
> https://github.com/cspr-rad/kairos/blob/2ef83df22bbdfa06f2d49b8fed309c0274ba907d/kairos-server/src/routes/transfer.rs#L14-L25
> 
> ➡️ There is no signing mechanism yet, so I would not worry about losing information or increasing complexity here. However, with #30 (already approved by Avi) we will have payload structure that will contain signature embedded **inside**, and no one will be interacting with signature directly - always via payload structure.
> 
> 🌵 Regarding `type PublicKey`, it is currently used for two things:
> 
> 1. Handling transfer's request body.
>    ➡️  Same story as with the signature - it will become obsolete, once we integrate new payload structure.
> 2. Storing balances:
>    https://github.com/cspr-rad/kairos/blob/2ef83df22bbdfa06f2d49b8fed309c0274ba907d/kairos-server/src/state.rs#L16-L17
>    
>    ➡️ If we consider this is the only place where `PublicKey` newtype is used, then I am not happy with seeing it inside `kairos-server` crate. I believe it could be placed in `kairos-types` or `kairos-crypto`, but at this point using it seems to provide no value.
> 
> cc @quinn-dougherty

I'd like to add to this discussion that the top-down design approach is problematic when we don't know the what changes will occur in casper-types over the next few months. I think it makes sense to use either the casper_types::Key struct or Vec<u8> for now and be more explicit in a future release.

marijanp:
The arguments don't make any sense to me, it is way easier to achieve the same outcome of your PR by just writing:
```
type PublicKey = Vec<u8>;
type Signature = Vec<u8>;
```
And if it was a newtype:
```
enum PublicKey(pub Vec<u8>);
enum Signature(pub Vec<u8>);
```
Instead of changing every single occurence of PubKey and Signature.
And when you argue that there are that little occurences of Signature and PubKey for **now** then it doesn't mean that there won't be many more in the future.
Newtyping or aliasing amongst others serves the purpose that one can quickly change the type.

koxu1996:
@marijanp `kairos-server` is obviously a prototype, not production ready code (there is more cleaning that needs to be done there). If you want to keep additional semantic of types that are not really used in the code and will get removed soon, then okay, let's do it your way. There is no point in wasting more time over such basic thing. Let's merge this PR finally.

marijanp (approved):

Avi-D-coder (approved):

PR #40, opened 2024-03-18
Casper Deploy Notifier

### LINKED ISSUE #45: SSE client for Casper
For deposits routed through Kairos, we should subscribe for Casper's `DeployProcessed` server-side event in order to get notification once deployment was processed. Relevant design part:

image

Notably there is no Rust client/SDK available for Casper's SSE.

koxu1996:
This PR implements basic SSE client for Casper, that looks for `DeployProcess` events. It will be used as a **additional trigger** for fetching deposit events data from blockchain state. However, we are not relying on it - it is just optimization for periodic polling.

You can read about Casper events here: https://docs.casper.network/developers/dapps/monitor-and-consume-events/.

Fixes #45. 

## Example usage

I provided `main.rs` that shows how `DeployNotifier` can be used. Here is screenshot of running binary for a minute:

image

Corresponding deployments can be located in block explorer:

image

image


xcthulhu:
What is the strategy for when the SSE drops a message?

koxu1996:
> What is the strategy for when the SSE drops a message?

@xcthulhu SSE gives us the information about deployment execution result, once they are included in the block. There is no information about related smart contract , so it it useful only for **known deployment hashes**. Since we have to process deployments that happened directly on L1 - without our prior knowledge - there is still need for additional component in Kairos that will periodically synchronize the state (I will be working on it next week). So SSE is just additional trigger for us synchronizing state, instead of waiting for next interval. Therefore we shouldn't care about dropped messages.

casper-deploy-notifier/src/types.rs
```diff
@@ -0,0 +1,6 @@
+#[derive(Debug)]
+pub struct Notification {
+    pub deploy_hash: String,
+    pub public_key: String,
+    pub success: bool,
+}
```
marijanp, line 6: I would suggest to factor out the `DeployProcessed` enum from the `SseData` enum and either use that type directly, or implement the `From` trait

casper-deploy-notifier/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+use crate::error::SseError;
+use crate::sse_types::SseData;
+use crate::types::Notification;
+
+mod error;
+mod sse_types;
+mod types;
+
+const DEFAULT_SSE_SERVER: &str = "https://events.mainnet.casperlabs.io";
+const DEFAULT_EVENT_CHANNEL: &str = "/events/main";
```
marijanp, line 15: No need to make those globals, if they are only used in the `Default` trait implementation.

casper-deploy-notifier/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+
+type BoxedEventStream = BoxStream<'static, Result<Event, EventStreamError<reqwest::Error>>>;
+
+pub struct DeployNotifier {
+    url: String,
```
marijanp, line 20: use reqwest's URL type for more type safety

casper-deploy-notifier/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+        let _api_version = match handshake_data {
+            SseData::ApiVersion(v) => Ok(v),
+            _ => Err(SseError::InvalidHandshake),
+        }?;
+
+        // Wrap stream with box.
+        let boxed_event_stream = Box::pin(event_stream);
+
+        Ok(boxed_event_stream)
+    }
```
marijanp, line 60: This does not need to be an associated function, it can be completely decoupled from the type and simply be a function. It only uses the `DeployNotifiers` url and does not need anything else to establish a connection.

casper-deploy-notifier/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+
+    // Handle incoming events - look for successfuly processed deployments.
+    pub async fn run(&mut self, tx: mpsc::Sender<Notification>) -> Result<(), SseError> {
+        let mut event_stream = self.connect().await?;
+
+        while let Some(event) = event_stream.try_next().await? {
+            let data: SseData = serde_json::from_str(&event.data)?;
+            match data {
+                SseData::ApiVersion(_) => Err(SseError::UnexpectedHandshake)?,
```
marijanp, line 69: I don't fully understand why this maps to an `UnexpectedHandshake` error.

casper-deploy-notifier/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+                SseData::DeployProcessed {
+                    execution_result,
+                    deploy_hash,
+                    account,
+                } => {
+                    let success = match *execution_result {
+                        casper_types::ExecutionResult::Failure { .. } => false,
+                        casper_types::ExecutionResult::Success { .. } => true,
+                    };
+                    let notification = Notification {
```
marijanp, line 80: This is where a `From` trait implementation would make the code much more simpler.

casper-deploy-notifier/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+        }?;
+
+        // Wrap stream with box.
+        let boxed_event_stream = Box::pin(event_stream);
+
+        Ok(boxed_event_stream)
+    }
+
+    // Handle incoming events - look for successfuly processed deployments.
+    pub async fn run(&mut self, tx: mpsc::Sender<Notification>) -> Result<(), SseError> {
```
marijanp, line 63: I would rename it to `subscribe`

casper-deploy-notifier/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+mod error;
+mod sse_types;
+mod types;
+
+const DEFAULT_SSE_SERVER: &str = "https://events.mainnet.casperlabs.io";
+const DEFAULT_EVENT_CHANNEL: &str = "/events/main";
+
+type BoxedEventStream = BoxStream<'static, Result<Event, EventStreamError<reqwest::Error>>>;
+
+pub struct DeployNotifier {
```
marijanp, line 19: I don't see a point why this even needs to be a type. A single function:
```Rust
async pub fn subscribe_to_node(url: &reqwest::URL, sender: mpsc::Sender<DeployProcessed>) {
  let stream = connect(url);
  ...
  sender.send(deploy_processed)
  // or if you want to continue to use Notification: sender.send(deploy_processed.into())
  ...
}
```

marijanp (commented):
I don't see a point why `DeployNotifier` needs to be a type. It's mapping `DeployProcessed` events comming from a node URL to a `Notification` and sends that through a channel.
A single function:
```Rust
async pub fn subscribe_to_node(url: &reqwest::URL, sender: mpsc::Sender<DeployProcessed>) {
  let stream = connect(url);
  ...
  sender.send(deploy_processed)
  // or if you want to continue to use Notification: sender.send(deploy_processed.into())
  ...
}
```
should be sufficient.

casper-deploy-notifier/src/types.rs
koxu1996: Done in https://github.com/cspr-rad/kairos/pull/40/commits/9372b7dd220ed0ec5919cc077c7916436e970654.

casper-deploy-notifier/src/lib.rs
koxu1996: `ApiVersion` is the first event that sent by server, that is used as the handshake. Once you received it you should not receive it again, thus `UnexpectedHandshake`.

I added clarification in https://github.com/cspr-rad/kairos/pull/40/commits/668faadaecc2a2492409beef8bbd0a2e19b96e4d.

casper-deploy-notifier/src/lib.rs
koxu1996: Done in https://github.com/cspr-rad/kairos/pull/40/commits/9372b7dd220ed0ec5919cc077c7916436e970654.

casper-deploy-notifier/src/lib.rs
koxu1996: Full SSE endpoint is composed of different 2 parts:
- server address - could be replaced in the future,
- event channel - always the same, as it is [Casper specific](https://github.com/casper-network/casper-node/blob/9f3995853204a18f17de9c022233d22aa14b9c37/node/src/components/event_stream_server/sse_server.rs#L45-L53).

Using those constants gives better semantic; also I prefer to see hardcoded values in the top of the file, even if they are used once.

casper-deploy-notifier/src/lib.rs
koxu1996: Maybe I am biased about `subscribe`, but just take a look at API of Casper SDKs:

```
watcher.subscribe(deploy_a);
watcher.subscribe(deploy_b);
watcher.subscribe(deploy_c);
let result = watcher.start().await; // runs until short timeout is reached, or event about deploy_a/b/c is received.
```

Therefore I think `run` is better for this long running function.

casper-deploy-notifier/src/lib.rs
koxu1996: Not possible if we want to provide `impl Default`.

casper-deploy-notifier/src/lib.rs
koxu1996: It logically belongs to `DeployNotifier`, but due to ownership it was easier to pass it via return value, rather than store it as part of state. Anyway, I think I managed to do it better now - https://github.com/cspr-rad/kairos/pull/40/commits/6f77813dea7d81bff87d158f24d68f492ee43f92.

casper-deploy-notifier/src/lib.rs
koxu1996: I could convert `DeployNotifier` to a single function now, but I prefer to have code prepared for future improvements, e.g. filtering, queuing, timeouts.

Avi-D-coder: I like it as a method

casper-deploy-notifier/src/lib.rs
marijanp: I don't think we should strive to follow the Casper SDK design decisions, as they seem to result in imperative and non-composable code.
The current API
```rust
let notifier = DeployNotifier { url, event_stream: None };
notifier.connect(); // mutates the DeployNotifier to now be conneced.
notifier.run(notification_sender);
```
Suggestion:
```rust
subscribe("http://casper-node.com", notification_sender);
```

casper-deploy-notifier/src/lib.rs
marijanp: This just means that there should be no `Default` implementation for `DeployNotifier` how can a `DeployNotifier` listen to anything without knowing about a url. there should be no notifier if there is no URL it can listen to.

casper-deploy-notifier/src/lib.rs
marijanp: What do you mean with method? an associated function?
Can you explain why you prefer a type here?

jonas089 (approved):
I have reviewed the changes and agree with them.
For Kairos we could implement an async logging loop similar to what Andrzej is demonstrating in echo_demo.rs.
Looks straight forward to use and clean. 
I am approving since I don't see a need to suggest changes.

casper-deploy-notifier/src/lib.rs
koxu1996: There is no way to construct complex `reqwest::Url` structure - that unnecessarily exposes internal implementation - without throwing error, and that is reason to not provide `Default`? You think about strict Kairos usage, however from the beginning Casper Deploy Notifier was aimed to be easy to use library that could be shared with community. Most users are expected to rely on default URL instead of providing custom one, as there is lack of public infrastructure in Casper.

casper-deploy-notifier/src/lib.rs
koxu1996: Method is associated function.

Too much abstraction can be overkill sometimes, but it does not hurt us, especially with such simple struct that is promising to evolve. Again, think about Casper community, not just Kairos.

Avi-D-coder: Both could work fine for now, I don' really care which, but I expect this to grow into the builder pattern.

casper-deploy-notifier/src/lib.rs
marijanp: You can call `parse` and unwrap if you are sure that the default URL you gave it is a valid URL, and still use `Default`. I think you misunderstand my comments. I'm trying to make suggestions towards a more simple API avoiding introducing state that is not obvious to the user.
As I said earlier:
```Rust
let notifier = default(); // is it listening yet, what does it listen to?
notifier.connect(); // what is it connecting to?
notifier.run(tx); // do I need to connect manually before or when it runs, do I need to disconnect manually?
```
```Rust
let notifier = new("http://test.com"); // is it listening yet?
notifier.run(tx); // do I need to connect manually before or when it runs, do I also need to disconnect manually?
```
```Rust
subscribe("http://test.com", sender); // no more work I need to do
```

Avi-D-coder (approved):
I'm good with this

Avi-D-coder (approved):
Lgtm

marijanp (approved):
Approving, but noting that there are no tests

koxu1996:
> Approving, but noting that there are no tests

Follow up task created - https://github.com/cspr-rad/kairos/issues/85.

PR #46, opened 2024-03-20
Add cctl module and extend e2e test to use it

marijanp:
- added a binary which uses the kairos-test-utils to launch a cctl network
- added a nixos module definition for cctl
- added the module to the e2e-tests server configuration, and created an assertion that queries one of the casper-nodes using casper-client-rs


nixos/modules/cctl.nix
```diff
@@ -0,0 +1,69 @@
...
+      type = types.package;
+    };
+
+    port = mkOption {
+      type = types.port;
+      default = 11101;
+      example = 60000;
+      description = mdDoc ''
+        Port to listen on.
+        TODO make port configurable in cctl
```
xcthulhu, line 28: Why not just do it now?

marijanp: Because CCTL does not offer a way to do that yet. Mark is aware of it and working on it.

Rom3dius:
Small question, looks all good to me and I think I'm good to approve.

How flexible are such nix modules? During the testing that @jonas089 did, he often had to modify the chainspec files in casper-node to allow them to accept the large proofs generated by the zk-rollup. Hopefully this won't be an issue with the trie, and if it is then wrapper proofs might be a solution anyway, but during this phase of prototyping such changes to the casper-node are practical.

marijanp:
> Small question, looks all good to me and I think I'm good to approve.
> 
> How flexible are such nix modules? During the testing that @jonas089 did, he often had to modify the chainspec files in casper-node to allow them to accept the large proofs generated by the zk-rollup. Hopefully this won't be an issue with the trie, and if it is then wrapper proofs might be a solution anyway, but during this phase of prototyping such changes to the casper-node are practical.

For different chainspec configurations you would extend the module with an according option.
You then implement this option, analogous to what I did with cctl `enable` or `port`. I assume given a chainspec option you create the respective updated chainspec and cctl is called with this chainspec.

> Also, shouldn't we be waiting at least for the genesis block before we start testing? `kairos.succeed("casper-client get-node-status --node-address http://localhost:11101")` Is it feasible to instead check that genesis or the first block has been processed before running tests?

The cctl rust-wrapper waits for that. So here we just wait for the service to start as expected

jonas089 (approved):

Rom3dius (approved):
Approved, but I don't think changing chainspec or other files will be easy using nix, since (as I understood from your last message) it seems mostly capable of flexibility in CLI arguments, but not in changing files (CCTL has no CLI argument for chainspec)

nixos/modules/cctl.nix
```diff
@@ -0,0 +1,69 @@
...
+    };
+
+    logLevel = mkOption {
+      type = types.enum [
+        "error"
+        "warn"
+        "info"
+        "debug"
+        "trace"
+      ];
```
Avi-D-coder, line 39: Can we get a string option for "RUST_LOG".
https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#directives

Avi-D-coder (approved):

nixos/modules/cctl.nix
marijanp: done [a0d71e8](https://github.com/cspr-rad/kairos/pull/46/commits/a0d71e8caa93ebe04a9c156e8b1ec749478d3b26)

marijanp:
@Rom3dius the option I meant in my explanation is a module option. How you implement a module option can differ. It can be whatever you want. So I could modify a file, set an env-var, pass an argument, run a binary, etc. it doesn't matter

Avi-D-coder (approved):
LGTM

koxu1996:
@Rom3dius Agree, this might be blocker for us when we start integrating Risc0. However, we should think about workarounds to avoid chainspec customization, as this prevents us from using official L1 infrastructure.

PR #49, opened 2024-03-22
Add kairos server test hook

marijanp:


kairos-test-utils/Cargo.toml
```diff
@@ -16,4 +16,5 @@ nom = "7"
 tokio = { version = "1", features = [ "full", "tracing", "macros" ] }
 tempfile = "3"
 tracing = "0.1"
+reqwest = { version = "*", features = ["json"] }
```
jonas089, line 19: What do we gain from version="*" here? I don't think it matters but I would like to understand :).

jonas089 (approved):

flake.nix
```diff
@@ -41,7 +41,7 @@
       ];
       perSystem = { config, self', inputs', system, pkgs, lib, ... }:
         let
-          rustToolchain = inputs'.fenix.packages.stable.toolchain;
+          rustToolchain = inputs'.fenix.packages.complete.toolchain;
```
jonas089, line 44: Shouldn't this be latest -> same problem we had with the deposit contract?

jonas089 (changes requested):

flake.nix
jonas089, line 44:
```suggestion
          rustToolchain = inputs'.fenix.packages.latest.toolchain;
```

jonas089 (changes requested):

flake.nix
koxu1996: `latest` is less stable than `complete`, which is also a nightly toolchain. I would use `stable`.

kairos-test-utils/Cargo.toml
koxu1996: There should not be bare wildcard here.

kairos-test-utils/src/lib.rs
```diff
@@ -1 +1,3 @@
+#![feature(async_closure)]
```
koxu1996, line 1:
```suggestion
```

I removed this line, and it successfully compiles with the latest stable. Are you sure this feature is necessary?

flake.nix
```diff
@@ -41,7 +41,7 @@
       ];
       perSystem = { config, self', inputs', system, pkgs, lib, ... }:
         let
-          rustToolchain = inputs'.fenix.packages.stable.toolchain;
```
koxu1996, line 44: No need to drop stable toolchain.

kairos-test-utils/src/kairos.rs
```diff
@@ -0,0 +1,68 @@
...
+        Ok(Kairos {
+            url,
+            process_handle,
+        })
+    }
+}
+
+impl Drop for Kairos {
+    fn drop(&mut self) {
+        let _ = self.process_handle.kill();
```
koxu1996, line 57: **Comment:** This is not a graceful shutdown, but for testing it seems to be okay.

kairos-test-utils/src/kairos.rs
```diff
@@ -0,0 +1,68 @@
...
+            .port()
+            .to_string();
+        let url = Url::parse(format!("http://127.0.0.1:{}", port).as_str()).unwrap();
+        let kairos = bin_dir().join("kairos-server");
+        let process_handle = Command::new(kairos)
+            .env("KAIROS_SERVER_PORT", &port)
+            .spawn()
+            .expect("Failed to start the kairos-server");
+
+        wait_for_port(url.socket_addrs(|| Option::None).unwrap().first().unwrap())
```
koxu1996, line 44: **Comment:** In the future we should introduce healtcheck mechanism - for example `/health` endpoint - because open port does not mean that server is ready to process requests.

koxu1996 (changes requested):
Please go back to stable Rust, then I have no other concerns about this PR.

kairos-test-utils/Cargo.toml
marijanp: I allow cargo to figure out the right version. The build/test determines whether my code continues to be correct. I don't have to pin semantic versions. If for some reason I run `cargo update` and a newer reqwest version is selected I fix the errors.

marijanp: Pinning a specific version just causes multiple package versions being locked by cargo instead of allowing cargo to figure out a common base.

Avi-D-coder: I don’t mind loose version bounds in an app or in this case a test lib, but my preference would be for most significant non zero SemVer component.

So in this case I'd prefer “0.12”.

marijanp: I require `0.12` now [c585c51](https://github.com/cspr-rad/kairos/pull/49/commits/c585c512f7db2e34d0c987ddbaccaa3350c825cf)

flake.nix
jonas089: @marijanp Please take a look at this, I remember that we were having issues with `complete` ?

marijanp: That was due to a version mismatch. What we did is merge a `latest` rust-std for wasm with a `stable` toolchain. I reverted this to use stable. See the passing pipeline.

Avi-D-coder (approved):

flake.nix
koxu1996: I guess you locally used that unstable feature, then applied [this simple fix](https://github.com/rust-lang/rust/issues/62290#issuecomment-1722085100), but forgot to switch back to stable.

koxu1996 (approved):

jonas089 (approved):

PR #50, opened 2024-03-26
Decouple transaction body from handlers

koxu1996:
This PR reduces complexity of integrating Tx format (#30) with server, by decoupling request body from handler and related helpers.


Avi-D-coder (approved):
This code is getting replaced shortly, so i don’t have strong feelings about this pr.

jonas089 (approved):
I don't see any problems with this PR and +1 Avi.

PR #51, opened 2024-03-26
Improved version of Deposit contract

jonas089:
+ Error handling
+ Segregation
- Create_Purse entry point -> now automatically setup upon contract installation
+ Access Control using SecurityBadges
- Primitive Access Control

This deposit smart contract targets the release branch of the official casper node repository, e.g. casper-network/release-1.5.6.

Contract types were introduced to write a serialized Deposit to the contract's event storage.


marijanp (commented):
Some first comments

contracts/contract-types/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "contract-types"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+serde = "1.0.197"
+casper-types = {git = "https://github.com/casper-network/casper-node", branch="release-1.5.6", default-features=false}
```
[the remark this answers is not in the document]
jonas089: It could be, but then it would point to casper types v 4.0.0, I'd rather have it all point to the temporary build target that is node v 1.5.6.

contracts/deposit-contract-tests/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dev-dependencies]
+base64 = { version = "0.21.2", default-features = false, features = ["alloc"] }
+casper-engine-test-support = {git = "https://github.com/casper-network/casper-node", branch="release-1.5.6", default-features=false}
+casper-execution-engine = {git = "https://github.com/casper-network/casper-node", branch="release-1.5.6", default-features=false}
+casper-contract = {git = "https://github.com/casper-network/casper-node", branch="release-1.5.6", default-features=false}
+casper-types = {git = "https://github.com/casper-network/casper-node", branch="release-1.5.6", default-features=false}
```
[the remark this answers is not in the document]
jonas089: What if the casper-node version on crates.io is updated?

contracts/contract-types/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "contract-types"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+serde = "1.0.197"
```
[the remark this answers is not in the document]
jonas089: Why?

contracts/deposit-contract-tests/Cargo.toml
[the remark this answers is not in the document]
marijanp: There is a lockfile, and you will still specify: `casper-node = "1.5.6`

contracts/deposit-contracts/contract/src/detail.rs
```diff
@@ -0,0 +1,110 @@
...
+pub fn get_optional_named_arg_with_user_errors<T: FromBytes>(
+    name: &str,
+    invalid: DepositError,
+) -> Option<T> {
+    match get_named_arg_with_user_errors::<T>(name, DepositError::Phantom, invalid) {
+        Ok(val) => Some(val),
+        Err(_) => None,
+    }
+}
+
```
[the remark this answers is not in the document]
jonas089: What I meant by the comment at the top of the detail.rs file is that I am making use of these functions because they were used in cep contracts. I didn't not write them and have no intention to touch them unless something is wrong / causing a problem.

contracts/contract-types/Cargo.toml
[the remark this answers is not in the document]
marijanp: cargo will have more flexibility resolving the dependency automatically given the workspace deps. If other crates you depend on lock a different version you will  need to get back and manually adjust this. the same thing applies to users of your crate.

contracts/deposit-contract-tests/Cargo.toml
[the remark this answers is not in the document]
jonas089: Do we commit the lockfile?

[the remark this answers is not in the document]
marijanp: Lockfiles should be commited, how else are we going to ensure reproducibility?

contracts/deposit-contracts/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+  "malicious-session",
+  "withdrawal-session",
+  "deposit-session",
+  "contract"
+]
+
+[workspace.package]
+name = "deposit-contracts"
+version = "0.1.0"
+edition = "2021"
```
marijanp, line 15:
```suggestion
edition = "2021"
license = "MIT OR Apache-2.0"
```

contracts/deposit-contracts/contract/src/detail.rs
```diff
@@ -0,0 +1,110 @@
...
+            Key::from(account_hash)
+        }
+        CallStackElement::StoredContract {
+            contract_package_hash,
+            ..
+        } => Key::from(contract_package_hash),
+    }
+}
+
+/// Traverse the callstack
```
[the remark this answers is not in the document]
jonas089: You are not supposed to use this function at all.

contracts/deposit-contracts/contract/src/detail.rs
```diff
@@ -0,0 +1,110 @@
...
+            api_error::result_from(ret).map(|_| data)
+        };
+        // Assumed to be safe as `get_named_arg_size` checks the argument already
+        res.unwrap_or_revert_with(DepositError::FailedToGetArgBytes)
+    } else {
+        // Avoids allocation with 0 bytes and a call to get_named_arg
+        Vec::new()
+    };
+
+    bytesrepr::deserialize(arg_bytes).map_err(|_| invalid)
```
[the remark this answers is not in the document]
jonas089: This will propagate a Casper APIError to the contract runtime. I did not write any of the code in detail.rs and it was used by Casper Labs in the cep standards.

contracts/deposit-contracts/contract/src/detail.rs
```diff
@@ -0,0 +1,110 @@
...
+}
+
+/// Read optional arg and propagate errors
+pub fn get_named_arg_with_user_errors<T: FromBytes>(
+    name: &str,
+    missing: DepositError,
+    invalid: DepositError,
+) -> Result<T, DepositError> {
+    let arg_size = get_named_arg_size(name).ok_or(missing)?;
+    let arg_bytes = if arg_size > 0 {
```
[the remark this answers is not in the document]
jonas089: Please refer to the other answers regarding comments on detail.rs.

contracts/deposit-contract-tests/.env
```diff
@@ -0,0 +1 @@
+PATH_TO_WASM_BINARIES="/Users/chef/Desktop/deposit/kairos/contracts/deposit-contracts/target/wasm32-unknown-unknown/release"
```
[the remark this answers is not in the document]
koxu1996: +1

contracts/deposit-contract-tests/Cargo.toml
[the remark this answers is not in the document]
koxu1996: Btw. for years official Rust guides were recommending to NOT commit lock files for libraries (only for applications), however this changed year ago - see https://github.com/rust-lang/cargo/issues/8728.

Now it is agreed that lock files should be always stored in VCS, but simply not published in libraries artifacts.

contracts/deposit-contracts/contract/src/detail.rs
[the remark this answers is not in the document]
koxu1996: I don't understand why `get_optional_named_arg_with_user_errors()` is useful.

Besides, there are many concerns about CEP-78 quality, so I would not just copy their code without further considerations.

contracts/deposit-contracts/contract/src/detail.rs
```diff
@@ -0,0 +1,110 @@
...
+        Ok(val) => Some(val),
+        Err(_) => None,
+    }
+}
+
+/// Read optional arg and propagate errors
+pub fn get_named_arg_with_user_errors<T: FromBytes>(
+    name: &str,
+    missing: DepositError,
+    invalid: DepositError,
```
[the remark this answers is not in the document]
koxu1996: It's just another CEP-78 thing. IMO there is no need for passing those error types.

contracts/deposit-contracts/readme.md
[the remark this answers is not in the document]
jonas089: will remove it for now.

contracts/deposit-contracts/malicious-session/src/main.rs
```diff
@@ -0,0 +1,15 @@
+#![no_std]
+#![no_main]
+use casper_contract::contract_api::{account, runtime, system};
+use casper_types::{runtime_args, ContractHash, RuntimeArgs, URef, U512};
+
```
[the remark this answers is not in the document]
jonas089: ok.

contracts/deposit-contracts/deposit-session/src/main.rs
```diff
@@ -0,0 +1,24 @@
...
+use casper_types::{runtime_args, ContractHash, RuntimeArgs, URef, U512};
+
+#[no_mangle]
+pub extern "C" fn call() {
+    let contract_hash: ContractHash = runtime::get_named_arg("deposit_contract");
+    let amount: U512 = runtime::get_named_arg("amount");
+    let source: URef = account::get_main_purse();
+    // create a temporary purse that can be passed to the deposit contract
+    let temp_purse: URef = system::create_purse();
+    // fund the temporary purse
```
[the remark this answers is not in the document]
jonas089: I think this improves readability.

contracts/deposit-contracts/contract/src/detail.rs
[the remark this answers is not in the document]
koxu1996: It MIGHT be propagated to ApiError, but it does not have to - depending of what called will do with error. So far I am not worried about losing error details, but rather about this error not being used in the code at all - is arguments handling done properly?

contracts/deposit-contracts/contract/src/security.rs
```diff
@@ -0,0 +1,80 @@
...
+        Ok((
+            match bytes[0] {
+                0 => SecurityBadge::Admin,
+                _ => return Err(bytesrepr::Error::LeftOverBytes),
+            },
+            &[],
+        ))
+    }
+}
+
```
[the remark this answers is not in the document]
jonas089: Is quite obvious but ok.

contracts/deposit-contracts/contract/src/security.rs
```diff
@@ -0,0 +1,80 @@
...
+pub fn sec_check(allowed_badge_list: Vec<SecurityBadge>) {
```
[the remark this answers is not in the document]
jonas089: I can rename the function, but I view this comment as a personal preference rather than an improvement.

contracts/deposit-contracts/contract/src/detail.rs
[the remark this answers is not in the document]
koxu1996: Here is part from Casper [documentation](https://docs.casper.network/concepts/callstack/):

> Developers can access the call stack with the [contract_api::runtime::get_call_stack](https://docs.rs/casper-contract/3.0.0/casper_contract/contract_api/runtime/fn.get_call_stack.html) function.
> 
> If session code calls a contract, which in turn calls another contract, then the session code would represent the zeroth entity in the stack, the contract called by the initiating session code would be the first and the contract called by the first contract would be the second.

I think this function is supposed to get **immediate caller**, so effectively _parent_ of current execution context (because we have session code that calls deposit contract). Having explanation in the code would be nice though.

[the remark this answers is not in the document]
jonas089: I have tested the get immediate caller function many times and it works well in contract context. I only use it in contract entry points so it will retrieve the n-1th entity calling the entry point.

contracts/contract-types/src/lib.rs
koxu1996, line 1: This code is not used in deposit contract, please remove it.

contracts/deposit-contracts/contract/src/security.rs
```diff
@@ -0,0 +1,80 @@
...
+        .unwrap_or_revert();
+    if !allowed_badge_list.contains(
+        &dictionary_get::<SecurityBadge>(get_uref(SECURITY_BADGES), &base64::encode(caller))
+            .unwrap_or_revert()
+            .unwrap_or_revert_with(DepositError::InsufficientRights),
+    ) {
+        revert(DepositError::InsufficientRights)
+    }
+}
+
```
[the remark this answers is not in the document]
jonas089: "When it should be used" doesn't make sense in this context since none of this functionality is exposed to the user. The user only needs to understand the entry points and how to call them.

contracts/deposit-contracts/contract/src/security.rs
```diff
@@ -0,0 +1,80 @@
...
+        storage::{dictionary_get, dictionary_put},
+    },
+    ext_ffi,
+    unwrap_or_revert::UnwrapOrRevert,
+};
+use casper_types::{
+    bytesrepr::{self, FromBytes, ToBytes},
+    CLTyped, Key,
+};
+
```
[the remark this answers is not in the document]
jonas089: There is no standard for this. This is an alternative to matching the caller to a value stored under a key. In cep-18 this was used to introduce different roles, one of which is the admin role. We can later also use it to blacklist accounts etc.

contracts/deposit-contracts/build-contracts.sh
koxu1996, line 1: Deposit contract should be included in the workspace, and build-able with `cargo build` (without additional bash scripts). This also introduced `wasm-opt` dependency that can be replaced with Rust library. This was already discussed - please see https://github.com/cspr-rad/kairos/pull/29#discussion_r1511331219.

contracts/deposit-contracts/contract/src/detail.rs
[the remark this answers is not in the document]
jonas089: Ok I will get rid of Cep-78 error handling if it bothers you, it doesn't break anything though and so long as the utilities don't cause issues I don't see why this is a problem.

Anyway, will simplify it.

contracts/deposit-contracts/contract/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+        last_processed_deposit_counter.into(),
+    );
+
+    let most_recent_deposit_counter = storage::new_uref(u64::from(0u8));
+    named_keys.insert(
+        KAIROS_MOST_RECENT_DEPOSIT_COUNTER.to_string(),
+        most_recent_deposit_counter.into(),
+    );
+
+    let (contract_hash, _) = storage::new_contract(
```
koxu1996, line 198: This allows to use Casper anti-feature of upgradeable contracts. What about creating locked contract?

contracts/deposit-contract-tests/src/test_fixture.rs
```diff
@@ -0,0 +1,244 @@
...
+            .get_expected_account(account)
+            .named_keys()
+            .get(name)
+            .expect("must have contract hash key as part of contract creation")
+            .into_hash()
+            .map(ContractHash::new)
+            .expect("must get contract hash")
+    }
+
+    pub fn install(&mut self, admin: AccountHash) {
```
koxu1996, line 99: Why we need admins for the contract? I believe there is no need for ACL, and it could be simplified a lot.

contracts/deposit-contracts/malicious-reader/src/main.rs
```diff
@@ -0,0 +1,12 @@
+#![no_std]
+#![no_main]
+use casper_contract::contract_api::{account, runtime, system};
+use casper_types::{RuntimeArgs, URef, U512};
+
```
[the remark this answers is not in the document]
koxu1996: I don't understand what is _malicious_ in terms of smart contracts. Is it about trying to call entrypoints that should be protected with ACL?

contracts/deposit-contract-tests/src/test_fixture.rs
jonas089: The admin will be able to withdraw from the deposit contract purse. I like the SecurityBadges.

contracts/deposit-contracts/withdrawal-session/src/main.rs
koxu1996, line 1: Withdrawals are not designed yet, so this should be removed form this PR.

contracts/deposit-contracts/malicious-reader/src/main.rs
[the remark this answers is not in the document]
jonas089: not only that, I also use it to ensure that the purse is returned with ADD access rights and can't be emptied by a malicious session.

koxu1996 (changes requested):
First pass

contracts/deposit-contracts/withdrawal-session/src/main.rs
jonas089: But ultimately it will be like this. There will be a withdrawal entry point called by a contract admin. The only alternative is multisig.

koxu1996: I heavily doubt it. Please do not make extra assumptions, and focus on deposit part.

contracts/deposit-contracts/contract/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+mod security;
+use security::{sec_check, SecurityBadge};
+mod entry_points;
+
+use contract_types::Deposit;
+
+#[no_mangle]
+pub extern "C" fn init() {
+    if runtime::get_key(KAIROS_DEPOSIT_PURSE).is_some() {
+        runtime::revert(DepositError::AlreadyInitialized);
```
marijanp, line 33: Will this return or abort the execution of `init`?

contracts/deposit-contracts/contract/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+    }
+    let security_badges_dict = storage::new_dictionary(SECURITY_BADGES).unwrap_or_revert();
+    storage::dictionary_put(
+        security_badges_dict,
+        &base64::encode(
+            Key::from(runtime::get_caller())
+                .to_bytes()
+                .unwrap_or_revert(),
+        ),
```
marijanp, line 42: could you assign a variable with a meaningful name and then pass that variable as an argument?

contracts/deposit-contracts/contract/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+        SecurityBadge::Admin,
+    );
+    let admin_list: Option<Vec<Key>> =
+        get_optional_named_arg_with_user_errors(ADMIN_LIST, DepositError::InvalidAdminList);
+    if let Some(admin_list) = admin_list {
+        for admin in admin_list {
+            storage::dictionary_put(
+                security_badges_dict,
+                &base64::encode(admin.to_bytes().unwrap_or_revert()),
```
marijanp, line 51: could you assign a variable with a meaningful name and then pass that variable as an argument?

contracts/deposit-contracts/contract/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+            Key::from(runtime::get_caller())
+                .to_bytes()
+                .unwrap_or_revert(),
+        ),
+        SecurityBadge::Admin,
+    );
+    let admin_list: Option<Vec<Key>> =
+        get_optional_named_arg_with_user_errors(ADMIN_LIST, DepositError::InvalidAdminList);
+    if let Some(admin_list) = admin_list {
+        for admin in admin_list {
```
marijanp, line 48: please add a small comment what this for loop does. Does it update the on-chain admin dictionary?

contracts/deposit-contracts/contract/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+        processed: false,
+    };
+
+    let kairos_deposit_event_dict_uref = runtime::get_key(KAIROS_DEPOSIT_EVENT_DICT)
+        .unwrap_or_revert_with(ApiError::MissingKey)
+        .into_uref()
+        .unwrap_or_revert();
+    storage::dictionary_put::<Vec<u8>>(
+        kairos_deposit_event_dict_uref,
+        &most_recent_deposit_counter_value.to_string(),
```
marijanp, line 103: Whoah, why do we need to convert the counter value to a string?

contracts/deposit-contracts/contract/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+        most_recent_deposit_counter_value,
+    );
+}
+
+#[no_mangle]
+pub extern "C" fn incr_last_processed_deposit_counter() {
+    sec_check(vec![SecurityBadge::Admin]);
+    let last_processed_deposit_counter_uref =
+        runtime::get_key(KAIROS_LAST_PROCESSED_DEPOSIT_COUNTER)
+            .unwrap_or_revert_with(ApiError::MissingKey)
```
marijanp, line 119: With all these error messages we could probably improve the verbosity by adding what key is missing

contracts/deposit-contracts/contract/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+        .unwrap_or_revert_with(ApiError::MissingKey)
+        .into_uref()
+        .unwrap_or_revert()
+        .with_access_rights(AccessRights::READ_ADD_WRITE);
+    system::transfer_from_purse_to_purse(deposit_purse, destination_purse, amount, None)
+        .unwrap_or_revert();
+}
+
+#[no_mangle]
+pub extern "C" fn change_security() {
```
marijanp, line 148: I would like to see documentation on when this will/should be called and what it does.

contracts/deposit-contracts/contract/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+    );
+    let contract_hash_key = Key::from(contract_hash);
+    runtime::put_key(KAIROS_DEPOSIT_CONTRACT_NAME, contract_hash_key);
+
+    // runtime arguments for contract initialization
+    let mut init_args = runtime_args! {};
+    if let Some(mut admin_list) = admin_list {
+        // prepare runtime arguments
+        init_args.insert(ADMIN_LIST, admin_list).unwrap_or_revert();
+    }
```
marijanp, line 212: Does this mean if there was no admin_list set, the contract won't have one at all? I would probably not allow this and fail

jonas089: No, the caller is the default admin.

jonas089: But Andrzej wants the access control to be deleted alongside the withdrawal EP. So if I end up doing that then there will be no admin at all.

contracts/deposit-contracts/contract/src/main.rs
jonas089: Yes that's something I can do. I actually thought about it before.

contracts/deposit-contract-tests/src/integration_tests.rs
```diff
@@ -0,0 +1,70 @@
...
+    fn should_install() {
+        let mut fixture: TestContext = TestContext::new();
+        fixture.install(fixture.account_1);
+    }
+
+    fn setup() -> (TestContext, AccountHash) {
+        let fixture: TestContext = TestContext::new();
+        let installer = fixture.account_1;
+        return (fixture, installer);
+    }
```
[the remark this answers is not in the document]
jonas089: The installer of the deposit contract is account_1, it doesn't "own" the contract.

e.g. installer = account_1 

I don't understand why you see a problem here, it is obvious that account_1 is the contract installer to anyone reading the code.

contracts/deposit-contracts/contract/src/main.rs
jonas089: Revert aborts.

contracts/deposit-contracts/contract/src/detail.rs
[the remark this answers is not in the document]
jonas089: ok.

contracts/deposit-contract-tests/.env
```diff
@@ -0,0 +1 @@
+PATH_TO_WASM_BINARIES="/Users/chef/Desktop/kairos/contracts/deposit-contracts/target/wasm32-unknown-unknown/release"
```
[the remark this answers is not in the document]
jonas089: ok.

contracts/deposit-contract-tests/src/test_fixture.rs
```diff
@@ -0,0 +1,244 @@
...
+    pub fn run_deposit_session(&mut self, amount: U512, account: AccountHash) {
+        let session_args = runtime_args! {
+            "amount" => amount,
+            "deposit_contract" => self.contract_hash("kairos_deposit_contract", account)
+        };
+        let session_request = ExecuteRequestBuilder::standard(
+            *DEFAULT_ACCOUNT_ADDR,
+            &format!(
+                "{}/{}",
+                *PATH_TO_WASM_BINARIES, "deposit-session-optimized.wasm"
```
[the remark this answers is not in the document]
jonas089: ok.

contracts/deposit-contracts/contract/src/main.rs
jonas089: Yes, it does.

contracts/deposit-contract-tests/src/integration_tests.rs
```diff
@@ -0,0 +1,70 @@
...
+    fn withdrawal_from_purse() {
+        let (mut fixture, installer) = setup();
+        fixture.install(installer);
+        fixture.run_deposit_session(U512::from(100000000000u64), installer);
+        fixture.run_withdrawal_session(fixture.account_1, U512::from(100000000000u64), installer)
+    }
+
+    // attack vector 1: session code attempts to empty purse
+    #[test]
+    fn run_malicious_session() {
```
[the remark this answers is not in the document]
jonas089: The deployment itself is expected to fail in the fixture.

contracts/deposit-contracts/contract/src/detail.rs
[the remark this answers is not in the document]
jonas089: I reviewed them and added a few comments. I am happy with them and think they are sound.

contracts/deposit-contract-tests/.env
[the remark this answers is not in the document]
jonas089: Remember that we had a call over this with Andrzej and how the execution engine resolves the path? I can remove my default value that points to stuff on my machine but ultimately this has to be an absolute path set by the user. If you have doubts then I encourage you and Andrzej to experiment with a relative path. Maybe you can get it to work but I wasn't able to and the only workaround for me was to change the execution engine code and point to a custom version of casper-node.

contracts/deposit-contracts/contract/src/main.rs
jonas089:
```
116 |     storage::dictionary_put::<Vec<u8>>(
    |     ---------------------------------- arguments to this function are incorrect
117 |         kairos_deposit_event_dict_uref,
118 |         &most_recent_deposit_counter_value,
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `&str`, found `&u64`
```

jonas089: Casper dict only accepts str as key.

contracts/contract-types/src/lib.rs
jonas089: Yes it is used in the deposit contract.

jonas089: use contract_types::Deposit;
...
```
    let new_deposit_record: Deposit = Deposit {
        account: get_immediate_caller().unwrap_or_revert(),
        amount,
        timestamp: None,
        processed: false,
    };
```

jonas089: It's necessary to have some kind of Deposit struct in the contract to store them as events.

contracts/deposit-contracts/deposit-session/src/main.rs
```diff
@@ -0,0 +1,24 @@
...
+#![no_main]
+use casper_contract::contract_api::{account, runtime, system};
+use casper_types::{runtime_args, ContractHash, RuntimeArgs, URef, U512};
+
+#[no_mangle]
+pub extern "C" fn call() {
+    let contract_hash: ContractHash = runtime::get_named_arg("deposit_contract");
+    let amount: U512 = runtime::get_named_arg("amount");
+    let source: URef = account::get_main_purse();
+    // create a temporary purse that can be passed to the deposit contract
```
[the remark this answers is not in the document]
jonas089: I added comments in two different places clarifying the need for a temporary purse. I hope this will help.

contracts/contract-types/src/lib.rs
jonas089: I will keep this conversation open for another day or so in case you want to add to this / in case there is another crate that exposes a deposit struct that could be used here. Should that not be the case then I will leave it as is for now and resolve this conversation.

contracts/deposit-contracts/build-contracts.sh
jonas089: Due to Marijans requirements for nixification it is beneficial to have the wasm32 build target crates in a seperate workspace. Should you have further questions you can discuss this subject with him.

contracts/deposit-contracts/contract/src/main.rs
jonas089: What would need to be changed to disable the upgradability?

contracts/deposit-contracts/contract/src/main.rs
jonas089: I'll declare it as event_dict_key o.e.

I agree that it isn't obvious.

contracts/deposit-contracts/contract/src/main.rs
```diff
@@ -0,0 +1,218 @@
...
+
+use contract_types::Deposit;
+
+// This entry point is called once when the contract is installed
+// and sets up the security badges with the installer as an admin or the
+// optional list of admins that was passed to the installation session as a runtime argument.
+// The contract purse will be created in contract context so that it is "owned" by the contract
+// instead of the installing account.
+#[no_mangle]
+pub extern "C" fn init() {
```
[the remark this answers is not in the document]
jonas089: The installer will by default always be an admin. If you pass a list that contains the installer it will be added only once / overriden. Should you not supply a list then there is only 1 admin. 

If you want to remove the admin from the list you will have to do that after the contract was installed. 
The way this is done is negotiable but a default admin is required because otherwise noone can udpate the list. The default admin could later be another contract or a multisig wallet, all of this will depend on our implementaiton.

Since I have removed all Withdrawal functionality from the contract as per Andrzej's request, we currently only need to restrict access to the "increase counter" function for the sliding window technique that was discussed during our design-sessions. The admin account will be responsible for updating the on-chain counter every time that a deposit was included in a batch / processed.

contracts/deposit-contract-tests/src/integration_tests.rs
```diff
@@ -0,0 +1,70 @@
...
+    fn setup() -> (TestContext, AccountHash) {
+        let fixture: TestContext = TestContext::new();
+        let installer = fixture.account_1;
+        return (fixture, installer);
+    }
+
+    #[test]
+    fn deposit_into_purse() {
+        let (mut fixture, installer) = setup();
+        fixture.install(installer);
```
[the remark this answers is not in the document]
jonas089: I have changed the test and now the installer is not the one depositing, instead an unrelated "user" account calls the deposit session.

contracts/deposit-contract-tests/src/test_fixture.rs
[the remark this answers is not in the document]
jonas089: I applied this change in a commit.

contracts/deposit-contracts/contract/src/main.rs
jonas089: Done.

contracts/deposit-contracts/malicious-reader/src/main.rs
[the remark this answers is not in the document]
koxu1996: That makes sense, thanks for explanation!

contracts/deposit-contracts/deposit-session/src/main.rs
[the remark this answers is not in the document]
koxu1996: @marijanp Currently, there isn't a native and secure method to pass tokens to smart contracts in Casper - see https://docs.casper.network/resources/tutorials/advanced/transfer-token-to-contract/. If you provide your purse address to the smart contract (as an argument for the entry point call), it gains full access and could potentially drain it. This is the reason an additional contract (session code) is employed to create a temporary purse.

contracts/deposit-contracts/contract/src/detail.rs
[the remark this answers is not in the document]
koxu1996: Notice that error is not used at all - you could replace `invalid` with `DepositError::Phantom` and it would still work. They got many things wrong, even simple things like getting optional args.

contracts/contract-types/src/lib.rs
koxu1996: Okay, I missed usage of it, because I was not expecting it will be used for events.

Regarding events itself, please use [Casper Event Standard](https://github.com/make-software/casper-event-standard). **This is really important**, as it allows block explorers and existing parsers to understand what was emitted by our contract.

contracts/deposit-contracts/build-contracts.sh
koxu1996: Nix is a great addition, but should not be a core requirement to build contracts. Especially that it can be done with extra crate that replaces custom bash script and extra host dependency.

contracts/deposit-contracts/contract/src/main.rs
koxu1996: @jonas089 It would be more helpful to get more context before conversation is resolved. Right now I have to go through every resolved conversation, then figure out what change you made.

From what I see in https://github.com/cspr-rad/kairos/pull/51/files#diff-53df1ae74d6a66c198d5f0fbabfcbda5d7525c472042d7b6260918827ebf5b28R211-R217 you disabled storing contract uref in current account. That's weird...

Please see `new_contract` [documentation](https://docs.rs/casper-contract/latest/casper_contract/contract_api/storage/fn.new_contract.html):

image

contracts/deposit-contract-tests/src/test_fixture.rs
koxu1996: This is big security risk, as it makes Kairos centralized solution. There MIGHT be a specific reason to include admins in L1 contract, but "liking" it is not enough to justify it.

contracts/deposit-contracts/contract/src/main.rs
koxu1996: @jonas089 I don't want to remove access control. I want to understand why you introduced it, and what are your intentions. Then we can talk whether it's necessary or not.

contracts/deposit-contract-tests/src/test_fixture.rs
jonas089: But this is exactly the design we discussed in our calls. We are planning to build a centralized demo with one single service/node. I don't think we changed the design spec since but if we did then nobody told me about it.

jonas089: I removed the withdrawal functionality since it is not part of the spec currently, but the admin updates the sliding window pointer. This was discussed in our flow chart sessions with Marijan and Avi.

jonas089: We could decide to get rid of the pointer / it could be stored on the centralized L2 server, but that would have to be discussed with the others since it deviates from our design.

We are not building a decentralized solution because we would need a decentralized sequencer for that. V0.1 of Kairos will be centralized and we have discussed this many times.

.gitignore
```diff
@@ -4,3 +4,4 @@ result
 target
 .tmp
 .nixos-test-history
+.env
```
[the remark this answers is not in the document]
Avi-D-coder: Not unless we move the defaults into code.

There are pros and cons to adding defaults in code.

Pro: it’s closer to usage, and an obvious place.
Con: you can now deploy a server in prod with settings you did not explicitly provide directly in your dockerfile or nix expression.

kairos-contracts/deposit-contract-tests/tests/test_fixture/mod.rs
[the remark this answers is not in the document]
jonas089: I personally prefer the concise style and wouldn't want to "have to write more". I see that you want the code to be more descriptive, but I don't think there is good enough of a reason to change the entire test implementation.

kairos-contracts/deposit-contract-tests/tests/test_fixture/mod.rs
```diff
@@ -0,0 +1,248 @@
...
+        )
+        .build();
+        self.builder
+            .exec(update_security_request)
+            .expect_failure()
+            .commit();
+    }
+
+    // see deposit-session
+    pub fn run_deposit_session(&mut self, amount: U512, installer: AccountHash, user: AccountHash) {
```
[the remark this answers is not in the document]
jonas089: ok.

kairos-contracts/deposit-contract-tests/tests/integration_tests.rs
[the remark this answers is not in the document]
jonas089: 1. The answer to question #1 is currently yes. I could add a restriction that there must be at least one admin, but then there would be no way of preventing a single invalid admin to be passed as an argument.

2. Valid case, I'll write tests for it.

contracts/deposit-contracts/withdrawal-session/src/main.rs
jonas089: Resolved because withdrawal has been removed.

kairos-contracts/deposit-contract-tests/tests/integration_tests.rs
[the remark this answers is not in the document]
jonas089: I added an additional test case for updating the last processed counter. The most recent counter has become obsolete due to CES.

kairos-contracts/deposit-contract-tests/tests/test_fixture/mod.rs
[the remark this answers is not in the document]
jonas089: Your other comment elaborates on this matter, I will resolve this conversation and we can continue the other discussion. I like the style of the test fixtures and if you want to re-implement the tests in a more functional style then that's a personal preference that's conflicting with mine.

kairos-contracts/deposit-contract-tests/tests/integration_tests.rs
```diff
@@ -0,0 +1,78 @@
...
+
+        let contract_balance_after = fixture.get_contract_purse_balance(installer);
+        assert_eq!(contract_balance_after, deposit_amount);
+
+        let user_balance_after = fixture.builder.get_purse_balance(user_purse_uref);
+        assert!(user_balance_after <= user_balance_before - deposit_amount);
+    }
+
+    #[test]
+    fn update_security_badges() {
```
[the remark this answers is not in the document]
jonas089: I just made the function name more descriptive in my last comment. Therefore I will resolve this conversation. You can unresolve it should you dislike the new, improved function name.

kairos-contracts/deposit-contracts/contract/src/utils.rs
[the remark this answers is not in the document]
jonas089: I improved the utils.rs and since they are only relevant for callstack traversal and reading the optional arguments, I would take them as they are at this point since the tests are passing and the traversal is sound.

kairos-contracts/deposit-contract-tests/tests/integration_tests.rs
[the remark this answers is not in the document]
jonas089: Please let me know if you find additional test cases that you would like me to cover. Let me know when we have reached a point where you wish to resolve this conversation.

contracts/deposit-contracts/contract/src/detail.rs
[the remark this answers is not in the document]
jonas089: I have changed this code since, therefore I am resolving this conversation since it points to an out-of-date commit.

contracts/contract-types/src/lib.rs
jonas089: CES is implemented now.

contracts/deposit-contracts/build-contracts.sh
jonas089: Everything is now buildable from the root of the kairos directory. Therefore I think the issue is resolved.

jonas089: Bash scripts are not needed anymore. Build process:

```
nix build .#kairos-contracts
cargo test -p deposit-contract-tests

```

jonas089: @koxu1996 Please let me know whether I can resolve this conversation.

contracts/deposit-contracts/contract/src/main.rs
jonas089: I just fixed this. Sorry, I was mistaken.

jonas089: In the latest commit I use locked_contract and am adding the package to the installers named keys

contracts/deposit-contracts/contract/src/detail.rs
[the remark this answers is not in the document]
jonas089: Error handling has been improved in the latest commit.

kairos-contracts/demo-contract/deposit-session/src/main.rs
```diff
@@ -0,0 +1,36 @@
...
+*/
+#![no_std]
+#![no_main]
+use casper_contract::contract_api::{account, runtime, system};
+use casper_types::{runtime_args, ContractHash, RuntimeArgs, URef, U512};
+
+#[no_mangle]
+pub extern "C" fn call() {
+    let contract_hash: ContractHash = runtime::get_named_arg("demo_contract");
+    let amount: U512 = runtime::get_named_arg("amount");
```
marijanp, line 17: I'm just wondering about the `constants` module which we are not using here: `RUNTIME_ARG_AMOUNT`

kairos-contracts/demo-contract/contract/src/utils/events.rs
```diff
@@ -0,0 +1,11 @@
+use casper_event_standard::Event;
+extern crate alloc;
+use alloc::string::String;
+use casper_types::{Key, U512};
+
+#[derive(Event)]
+pub struct Deposit {
+    pub account: Key,
+    pub amount: U512,
+    pub timestamp: Option<String>,
```
marijanp, line 10: I assume we have no other options for the timestamp type?

kairos-contracts/demo-contract/contract/src/main.rs
```diff
@@ -0,0 +1,130 @@
...
+        .unwrap_or_revert_with(DepositError::MissingKeyDepositPurse)
+        .into_uref()
+        .unwrap_or_revert_with(ApiError::UnexpectedKeyVariant);
+    system::transfer_from_purse_to_purse(temp_purse, deposit_purse_uref, amount, None)
+        .unwrap_or_revert();
+
+    let new_deposit_record: Deposit = Deposit {
+        account: get_immediate_caller().unwrap_or_revert(),
+        amount,
+        timestamp: None,
```
marijanp, line 96: Are we going to use the timestamp at all?

kairos-contracts/demo-contract/malicious-reader/src/main.rs
```diff
@@ -0,0 +1,19 @@
...
+#![no_main]
+use casper_contract::contract_api::{account, runtime, system};
+use casper_types::{URef, U512};
+
+#[no_mangle]
+pub extern "C" fn call() {
+    let amount: U512 = runtime::get_named_arg("amount");
+    let purse_uref: URef = runtime::get_named_arg("purse_uref");
+    let destination_purse: URef = account::get_main_purse();
+    system::transfer_from_purse_to_purse(purse_uref, destination_purse, amount, None)
```
marijanp, line 17: is this only expected to fail because `purse_uref.with_access_rights(AccessRights::WRITE);` is missing?

kairos-contracts/demo-contract-tests/tests/integration_tests.rs
```diff
@@ -0,0 +1,64 @@
...
+        let mut fixture: TestContext = TestContext::new();
+        fixture.install_demo_contract(fixture.account_1);
+    }
+
+    fn setup() -> (TestContext, AccountHash, AccountHash) {
+        let fixture: TestContext = TestContext::new();
+        let installer = fixture.account_1;
+        let user = fixture.account_2;
+        (fixture, installer, user)
+    }
```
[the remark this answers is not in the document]
marijanp: If `account_1` is not capturing the semantics for the test case correctly, then why not just rename the field from `account_1` to `installer`.

kairos-contracts/demo-contract-tests/tests/integration_tests.rs
```diff
@@ -0,0 +1,64 @@
+mod test_fixture;
+#[cfg(test)]
+mod tests {
+    use crate::test_fixture::TestContext;
+    use casper_types::{account::AccountHash, U512};
+
+    #[test]
+    fn should_install_deposit_contract() {
+        let mut fixture: TestContext = TestContext::new();
+        fixture.install_demo_contract(fixture.account_1);
```
marijanp, line 10: The fixture should encapsulate who is the installer. As in other comments, I don't see the benefit of introducing a fixture that does not encapsulate any internal knowledge. `account_1` is clearly used as the installer across all tests. It should just be called `installer`. the `new` associated function should then call `install_demo_contract()` without an argument, and just installing it using the `installer` account. Either encapsulate, or make it parametrized and modular.

kairos-l1-utils/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "kairos-l1-query"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+
+[dependencies]
+casper-client = {git="https://github.com/casper-ecosystem/casper-client-rs", branch="release-2.0.0"}
```
marijanp, line 8: Why are we using release 2.0.0

kairos-l1-utils/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+        casper_client::Verbosity::Low,
+        deploy,
+    )
+    .await
+    .unwrap()
+}
+
+pub async fn query_state_root_hash(node_address: &str, rpc_port: &str) -> Digest {
+    get_state_root_hash(
+        JsonRpcId::String(rpc_port.to_owned()),
```
marijanp, line 49: I'm pretty sure this is not the way the port should be used?

kairos-l1-utils/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+    let value: u64 = match stored_value {
+        StoredValue::CLValue(cl_value) => cl_value.into_t().unwrap(),
+        _ => panic!("Missing Value!"),
+    };
+    value
+}
+
+#[tokio::test]
+async fn state_root_hash() {
+    let srh = query_state_root_hash("http://127.0.0.1:11101/rpc", "11101").await;
```
marijanp, line 84: You should use `kairos-test-utils` and spawn a cctl network by adding:
```
 let network = CCTLNetwork::run().await.unwrap();
 let srh = query_state_root_hash(&format!("http://localhost:{}", node.port.rpc_port), "11101").await;
```

kairos-l1-utils/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+    println!("Srh: {:?}", &srh);
```
marijanp, line 85: A test usually asserts something

kairos-l1-utils/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+#[tokio::test]
+async fn install_wasm() {
+    use std::fs::File;
+    use std::io::Read;
+    use casper_types::{RuntimeArgs, runtime_args};
+    let mut wasm_file: File =
+        File::open("/Users/chef/Desktop/demo-contract-optimized.wasm").unwrap();
```
marijanp, line 94: This is an absolute path on your machine

marijanp (changes requested):

kairos-contracts/deposit-contract-tests/tests/test_fixture/mod.rs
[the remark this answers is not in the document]
marijanp: This is unfortunately not concise, but rather confusing.

kairos-l1-utils/Cargo.toml
jonas089: Because that is the latest stable release and generally Casper Client is pretty bad. The dev branch breaks when you try to import it as a library.

kairos-l1-utils/src/lib.rs
jonas089: It's pretty weird and doesn't seem to do anything... The way I am doing it works but I am also confused so I'll try to figure out how they intended it.

kairos-l1-utils/src/lib.rs
jonas089: Srh is not deterministic so this test either fails or succeeds.

kairos-l1-utils/src/lib.rs
jonas089: Does Kairos test utils include account management?

marijanp: What kind of account management do you need? the signing key can be obtained by using the `assets_dir` field

kairos-l1-utils/src/lib.rs
marijanp: "It works" is not reasonable. The rpc id is used for matching requests and responses as they are not guaranteed to arrive in a specific order. Please just use an arbitrary id, instead of mispurposing the port which is semantically wrong.

jonas089: I accidentally clicked on resolve earlier.

If it is an arbitrary ID that we don't need / use then what I'm currently doing will never cause a problem, we can replace it with "0" but does it matter?

marijanp: It does, because in your function you expect an `rpc_port` as an argument that you don't even use later on when you query the state root hash. You instead use it as an `rpc_id`. Anyone using or maintaining your code will be confused about the purpose this argument serves

kairos-l1-utils/Cargo.toml
jonas089: I'll remove it and default the id to 0

kairos-l1-utils/src/lib.rs
jonas089: I'll remove and default the id to 0

kairos-l1-utils/Cargo.toml
marijanp: 2.0.0 is available on crates.io, what revision do you need exactly and why?

kairos-contracts/demo-contract/deposit-session/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "deposit-session"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+
+[dependencies]
+casper-contract = {git="https://github.com/cspr-rad/casper-node", branch="kairos-demo-1.5.6", features=["test-support"]}
```
marijanp, line 8: Why are we picking this specific branch of `1.5.6`? why not upstream `1.5.9`

jonas089: Because I added the risc0 host function verifier to this branch.

jonas089: If we want to be able to verify the proof on the L1 then we have to use this branch that I created.

kairos-contracts/demo-contract/contract/src/utils/errors.rs
```diff
@@ -0,0 +1,24 @@
...
+    MissingKeyMostRecentDepositCounter = 6,
+    MissingKeyLastProcessedDepositCounter = 7,
+    MissingKeyDepositEventDict = 8,
+    FailedToCreateDepositDict = 9,
+    FailedToReturnContractPurseAsReference = 10,
+}
+
+impl From<DepositError> for ApiError {
+    fn from(error: DepositError) -> Self {
+        ApiError::User(error as u16)
```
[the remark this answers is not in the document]
jonas089: MissingArgument is one error. UserError takes a value and can represent any error not handled by the available set of verbose errors. This is standard practice on Casper.

kairos-contracts/demo-contract/deposit-session/Cargo.toml
marijanp: And why is it 1.5.6?

jonas089: It was 1.5.6 before. There is no 1.5.9, 1.5.6 is the latest stable release of 1.5.x.

Rom3dius (approved):
lgtm

kairos-contracts/demo-contract/deposit-session/Cargo.toml
Rom3dius: Resolve this, 1.5.6 is the latest stable release and what we all agreed to do the demo on

marijanp: Oh sorry about that, my fault

kairos-contracts/demo-contract/deposit-session/src/main.rs
jonas089: I chose not to make the main contract a lib, it would be a bit odd if the test sessions depended on the main contract.

kairos-contracts/demo-contract/contract/src/main.rs
jonas089: Good question, @Avi-D-coder, @koxu1996 what do you think, should I remove it?

kairos-contracts/demo-contract/malicious-reader/src/main.rs
jonas089: Yes, exactly.

kairos-contracts/demo-contract-tests/tests/integration_tests.rs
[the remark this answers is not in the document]
jonas089: Because any account can be an installer. If I want to use the account_x as the installer when writing tests then I can assign it as such:

let installer = fixture.account_x

contracts/deposit-contracts/contract/src/main.rs
koxu1996: Thanks, I see it's fixed:

image

Avi-D-coder:
hmm
```
kairos-tarpaulin> /build/.tmpFYXGAy/assets/users/user-1/secret_key.pem
kairos-tarpaulin> thread 'install_wasm' panicked at kairos-l1-utils/src/lib.rs:43:6:
kairos-tarpaulin> called `Result::unwrap()` on an `Err` value: ResponseIsRpcError { rpc_id: String("0"), rpc_method: "account_put_deploy", error: Error { code: -32008, message: "Invalid Deploy", data: Some(String("block chain has no blocks")) } }
kairos-tarpaulin> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
kairos-tarpaulin> failures:
kairos-tarpaulin>     install_wasm
```

jonas089:
It looks like it doesn't properly await the creation of Blocks on startup?

marijanp (approved):

contracts/deposit-contracts/build-contracts.sh
koxu1996: `wasm-opt` is still additional host dependency, but it can be fixed later.

.github/workflows/check.yml
```diff
@@ -51,6 +50,18 @@ jobs:
...
           path: coverage-result/cobertura.xml
           minimum_coverage: 60
+      
+      - name: kairos-contracts x86_64-linux
+        if: matrix.os == 'ubuntu-latest'
+        run: nix build -L --no-link --show-trace .#packages.x86_64-linux.kairos-contracts
+
+      - name: kairos-contracts x86_64-darwin
+        if: matrix.os == 'macos-latest'
+        run: nix build -L --no-link --show-trace .#packages.x86_64-darwin.kairos-contracts
```
Avi-D-coder, line 60: Let's not add back x86_64-darwin.
aarch64-darwin is enough
```suggestion
```

Avi-D-coder (approved):
LGTM

.github/workflows/check.yml
marijanp: Will be addressed in other PR as discussed.

koxu1996 (approved):
:rocket:

marijanp (approved):

.github/workflows/check.yml
```diff
@@ -1,7 +1,6 @@
 name: check
 on:
   pull_request:
-    branches: [main]
```
marijanp, line 4: Somehow this sneaked in

PR #56, opened 2024-03-29
ci: add cargo test reporting

marijanp:


flake.nix
```diff
@@ -41,7 +41,7 @@
       ];
       perSystem = { config, self', inputs', system, pkgs, lib, ... }:
         let
-          rustToolchain = inputs'.fenix.packages.stable.toolchain;
+          rustToolchain = inputs'.fenix.packages.latest.toolchain;
```
jonas089, line 44: Should this be "latest" or "complete"?

jonas089 (approved):

flake.nix
koxu1996, line 44: There is no need to use unstable toolchain here.

flake.nix
```diff
@@ -83,6 +83,11 @@
 
             kairos = craneLib.buildPackage (kairosNodeAttrs // {
               cargoArtifacts = self'.packages.kairos-deps;
+              cargoTestExtraArgs = "-- -Z unstable-options --format json --report-time | tee test-output.json";
```
koxu1996, line 86:
```suggestion
              cargoTestExtraArgs = "-- --format json --report-time | tee test-output.json";
```

Would this work?

.github/workflows/check.yml
```diff
@@ -54,7 +54,17 @@ jobs:
 
       - name: kairos x86_64-linux
         if: matrix.os == 'ubuntu-latest'
-        run: nix build -L --no-link --show-trace .#packages.x86_64-linux.kairos
+        run: nix build -L --out-link kairos-result --show-trace .#packages.x86_64-linux.kairos
+
+      - name: Create JUnit Kairos Test Report x86_64-linux
+        if: matrix.os == 'ubuntu-latest'
+        run: cat kairos-result/test-result.json | nix run nixpkgs#cargo2junit > kairos-test-result.xml
```
koxu1996, line 61: Using `cargo2junit` forces us to use nigthly toolchain :disappointed:.

What about using `cargo nextest` instead of default cargo test runner? Nextest has JUnit output and it was designed for usage in CI.

PR #72, opened 2024-04-18
flake: use casper-network/cctl

marijanp:
This also updates the casper-node dependency of `cctl` from `1.5.4` to `1.5.6`


flake.nix
```diff
@@ -27,8 +27,8 @@
     advisory-db.flake = false;
     risc0pkgs.url = "github:cspr-rad/risc0pkgs";
     risc0pkgs.inputs.nixpkgs.follows = "nixpkgs";
-    csprpkgs.url = "github:cspr-rad/csprpkgs/add-cctl";
-    csprpkgs.inputs.nixpkgs.follows = "nixpkgs";
+    cctl.url = "github:casper-network/cctl/fix-macos";
```
koxu1996, line 30: Why `fix-macos` branch? I see that there is still [concern about it failing on macOS](https://github.com/casper-network/cctl/pull/12#pullrequestreview-2034158483).

marijanp: The pre-installed `cp` command on MacOS does not support `--no-preserve=mode` in the Nix work I packaged it with GNU's `coreutils` which contains the proper `cp` command. Unfortunately there is no alternative for the MacOS binary. This is subject to discussions. If we don't use `--no-preserve=mode` we run into permission problems, since the chainspec files come from the Nix store, and those files are owned by root. It would be good to find a way to support both Nix and otherwise vendored tooling.

Avi-D-coder (approved):

flake.nix
marijanp: They are not only owned by root but also read-only. but we need write permissions too.

koxu1996: With recent `chmod` fix in https://github.com/casper-network/cctl/pull/12/commits/947c34b991e37476db82ccfa2bd7c0312c1a91d7 I am happy with changes.

koxu1996 (approved):

flake.nix
```diff
@@ -25,8 +25,8 @@
     crane.inputs.nixpkgs.follows = "nixpkgs";
     advisory-db.url = "github:rustsec/advisory-db";
     advisory-db.flake = false;
-    csprpkgs.url = "github:cspr-rad/csprpkgs/add-cctl";
-    csprpkgs.inputs.nixpkgs.follows = "nixpkgs";
+    cctl.url = "github:casper-network/cctl/fix-macos";
```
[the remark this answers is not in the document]
marijanp: That is a fair point, I'm referring to the revision instead of the branch now,

Avi-D-coder (approved):

asladeofgreen (commented):
LGTM

PR #57, opened 2024-04-02
Integrate the trie

Avi-D-coder:
This PR just sets up the trie thread and state.
We still need all the prior logic against the btree to pre-validate transactions.
That btree will be replaced by a cache db `(PublicKey -> Account)`.

We could remove pre-validated transaction requirement with an immutable `trie::SnapshotBuilder` that supported could rollback.
However, we would still need to prevent DoS with the cache db. Real hardware is very different from the zkVM, binary tries don't perform all that well on SSDs. 

In the future we will also need to warm up the `NodeHash -> TrieNode` DB right before a txn hits the trie state thread. By cashing `TrieNode`s into ram we can move slow IO off that thread.

Next Steps after this PR:
- Adding persistence with rocks
- Adding a validating thread for testing against (like a mock contract for testing)
- General testing txns (we can replicate what I did in the trie repo).

PR #81, opened 2024-04-29
Refactor `kairos-tx`

koxu1996:
I extracted great improvements of `kairos-tx` done by @Avi-D-coder in [Trie integration](https://github.com/cspr-rad/kairos/pull/57), to reduce clutter in his PR.

List of changes:
- Now all structures - `Deposit`, `Withdrawal`, `Transfer`, `SigningPayload` - can be created with flexible constructor. Thus `helpers.rs` got removed.
- Now DER encoding/decoding is part of `SigningPayload`.
- Updated Nix e2e test to use 2 different account for deposit and withdrawal, also fixed nonces.
- Added Rust e2e test for encoding payloads.

One additional change that I made myself is to allow cloning the base fields, so we can extract nonce/amount/public-key without taking ownership of full transaction - https://github.com/cspr-rad/kairos/pull/81/commits/9fc5a007295bbfdeff54da64dd57493821b2f9ff.


Rom3dius (approved):
Looks good to me. I'm sure I'm missing context here, but why the name "der_encode" and "der_decode"?

Also these github actions are out of control, take far to long to run.

koxu1996:
@Rom3dius Thanks! Let me provide you context :wink: [Schema for Kairos transaction](https://github.com/cspr-rad/kairos/blob/ac86f697e5c5e372a8dff1c0ebe252b68ac92193/kairos-tx/schema.asn) is defined as **ASN.1**:

```
TxSchema DEFINITIONS AUTOMATIC TAGS ::= BEGIN

  -- Basic types.
  PublicKey ::= OCTET STRING
  Amount ::= INTEGER (0..18446744073709551615)

  Transfer ::= SEQUENCE {
    recipient PublicKey,
    amount Amount,
    ...
  }
```

Notably it's fully abstract syntax - there is no definition how those should be converted from and to bytes. However, ASN.1 is closely associated with multiple sets of [encoding rules](https://en.wikipedia.org/wiki/ASN.1#Encodings): BER, DER, CER, and others. DER-encoding is the one we are using, as it is fully deterministic, so we can have the exact same bytes representation that is implementation agnostic.

Avi-D-coder (approved):
Lgtm

marijanp:
You did not address the comments I left about some of the changed files in Avi's PR:
- Removing the Nonce from deposits
- An explanation for why the payloads and signatures the way they changed in the end-to-end test

koxu1996:
@marijanp I believe that simple change in Nix e2e tests [was explained 3 days ago](https://github.com/cspr-rad/kairos/pull/57#discussion_r1581419918). You are usually not closing PR discussions that you started, so should I wait for you to click "resolve conversation"?

Your objectives about `nonce` in schema are not relevant here: this PR is basically refactoring API of schema, but not changing schema itself.

PR #84, opened 2024-04-29
flake: don't follow our nixpkgs revision for risc0pkgs to speedup builds

marijanp:
- I removed the `follows` instruction such that we trigger less rebuilds for the risc0 tooling if we update nixpkgs and benefit from better caching
- I updated the kairos-prover to follow the `risc0pkgs` template
- I packaged the kairos-prover with crane
- I refactored the source filtering for kairos, making it easier to use and likely willl improve caching because it's less error prone

Further temporary optimisation:
- uncomment the import of `./kairos-prover` to avoid building any risc0 related code
- removed the GHA ubuntu build
- removed the GHA `x86_64-darwin` build

Note that prior to merging we need to change the merge rules to not require the ubuntu-latest check.


Avi-D-coder (approved):

Avi-D-coder (approved):

Avi-D-coder (approved):

kairos-prover/rust-std-Cargo.lock
```diff
@@ -0,0 +1,6636 @@
+# This file is automatically @generated by Cargo.
```
koxu1996, line 1: **Question:** where is TOML file used to generate this lockfile?

marijanp: There is none because rust-std only ships with a lockfile: I don't know how you obtain your rust compiler but there should be a directory `lib/rustlib/src/rust` and this is where this fiel is copied from

PR #89, opened 2024-05-03
kairos-server: refactor config and main

marijanp:
- replaced port config option with socket addr
- refactored the main to solely parse the config and then run the server given a config


kairos-server/src/main.rs
```diff
@@ -1,21 +1,13 @@
...
         .unwrap_or_else(|e| panic!("Failed to parse server config from environment: {}", e));
-
-    let app = kairos_server::app_router(BatchStateManager::new_empty());
-
-    let axum_addr = SocketAddr::from(([127, 0, 0, 1], config.port));
-
-    tracing::info!("starting http server on `{}`", axum_addr);
-    let listener = tokio::net::TcpListener::bind(axum_addr).await.unwrap();
-    axum::serve(listener, app).await.unwrap();
+    run(config).await
```
Avi-D-coder, line 12: NIT: I'd prefer if we used the qualified path `kairos_server::run` for clarity.

Avi-D-coder (approved):

kairos-server/src/main.rs
```diff
@@ -1,21 +1,13 @@
-use std::net::SocketAddr;
-
 use dotenvy::dotenv;
-use kairos_server::{config::ServerConfig, state::BatchStateManager};
+use kairos_server::{config::ServerConfig, run};
 
 #[tokio::main]
 async fn main() {
-    tracing_subscriber::fmt::init();
```
koxu1996, line 8: **Suggestion:** Initiate the subscriber as the first action in `main()`, as we may need to emit logs before the server is actually run.

kairos-server/src/lib.rs
```diff
@@ -22,3 +25,15 @@ pub fn app_router(state: Arc<state::BatchStateManager>) -> Router {
...
+
+pub async fn run(config: ServerConfig) {
+    tracing_subscriber::fmt::init();
+
+    let app = app_router(BatchStateManager::new_empty());
+
+    tracing::info!("starting http server on `{}`", config.socket_addr);
+    let listener = tokio::net::TcpListener::bind(config.socket_addr)
+        .await
+        .unwrap();
```
koxu1996, line 37: **Suggestion:** Since we allow binding to port 0, it would be useful to log the address with bound port:

```suggestion
    let listener = tokio::net::TcpListener::bind(config.socket_addr)
        .await
        .unwrap();
    tracing::info!("listening on `{}`", listener.local_addr().unwrap());
```

Example logs for `KAIROS_SERVER_SOCKET_ADDR=0.0.0.0:0`:

```
2024-05-06T11:05:38.197541Z  INFO kairos_server: starting http server on `0.0.0.0:0`
2024-05-06T11:05:38.197591Z  INFO kairos_server: listening on `0.0.0.0:43653`
```

koxu1996 (approved):
Looks good, minor improvements suggested.

Rom3dius (approved):

kairos-server/src/main.rs
marijanp: There is not a lot happening in `main` only sourcing the environment which is ok to fail, and the parsing of the config, which panics. I understand that we might want to add logs in the future here when the `env` sourcing fails, but If we move it to main again we loose the nice property that `run` does everything we need to run the server with everything it needs. Logging initialization would need to be done additionally. However, we could consider calling `init` twice. in main and `run`.

kairos-server/src/main.rs
marijanp: done in [204ca19](https://github.com/cspr-rad/kairos/pull/89/commits/204ca19a4fc22b9bd8f69d8e9228ae58025663d1)

kairos-server/src/main.rs
koxu1996: If someone attempts to replace panics with logging in config parsing - an approach that seems quite likely - they MUST rework the subscriber initialization. However, it would be very easy to overlook this additional step and I would like to avoid creating this technical debt.

Where else is `run` called? If it's within `kairos-test-utils`, then I'm confident that we don't need logging there. Additionally, having the ability to configure subscribers externally seems like a desirable feature. I'm not very familiar with `tracing`; perhaps @Avi-D-coder could share his thoughts on this.

marijanp: The docs say that init panics if it's called twice: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/fn.init.html

`run` ins called in `kairos-test-utils` thus basically anywhere where tests might be run. We ideally want our server to produce logs when a test fails or during development.

Avi-D-coder: I prefer `tracing_subscriber::fmt::init()` at the top of main, and I want it enabled in tests.

But it's a minor thing so I don't care until we need it.

Avi-D-coder (approved):

koxu1996 (approved):

PR #90, opened 2024-05-03
kairos-test-utils/kairos: refactor to use kairos_server::run instead of binary

marijanp:
## Things done
- What the description says.
## Overview
#92 
## Motivation

Motivation is that we need this when testing the CLI, and this resolves a chicken-and-egg problem with the kairos-server cli

### DIFF
--- a/kairos-test-utils/src/kairos.rs
+++ b/kairos-test-utils/src/kairos.rs
@@ -25,6 +25,7 @@ impl Kairos {
         let config = kairos_server::config::ServerConfig { socket_addr };
 
         let process_handle = tokio::spawn(async move {
+            tracing_subscriber::fmt::init();
             kairos_server::run(config).await;
         });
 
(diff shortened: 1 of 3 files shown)


kairos-test-utils/src/kairos.rs
```diff
@@ -25,23 +14,21 @@ async fn wait_for_port(address: &SocketAddr) -> Result<(), io::Error> {
 
 pub struct Kairos {
     pub url: Url,
-    process_handle: Child,
+    process_handle: tokio::task::JoinHandle<()>,
```
Avi-D-coder, line 17: NIT: It'd be nice to name this `task_handle`.

Avi-D-coder (dismissed):

koxu1996 (dismissed):

Rom3dius (approved):

Issue #23, opened 2024-02-15
Autogenerate client SDKs


koxu1996:
Besides CLI for users, we want to provide set of SDKs for developers.

Since server is accessible via REST API, we can leverage OpenAPI spec to autogenerate clients (a.k.a. SDKs).

PR #91, opened 2024-05-03
kairos-cli: implement call to deposit endpoint and integration test

marijanp:
## Things done

- Moved all code except cli parsing to `kairos_cli::run`
- Introduced new `kairos-server-address` option, that defaults a value if not set
- Implemented a client function that will submit a transaction to the `kairos-server`, this client function introduces a `KairosClientError` enum type that could be used for better error handling
- implemented the mapping from cli arguments for a deposit -> deposit payload, signing and submitting of a deposit
- Updated the integration test to actually spawn a kairos-server
- Updated the ne2e-test

## Overview

#92 

## Motivation
-

### DIFF
--- a/kairos-cli/src/client.rs
+++ b/kairos-cli/src/client.rs
@@ -1,6 +1,6 @@
 use kairos_server::routes::PayloadBody;
 
-use reqwest::{Client, Url};
+use reqwest::{blocking, Url};
 use serde::{Deserialize, Serialize};
 use std::fmt;
 
@@ -36,18 +36,17 @@ impl From<reqwest::Error> for KairosClientError {
     }
 }
 
-pub async fn submit_transaction_request(
-    client: &Client,
+pub fn submit_transaction_request(
     base_url: &Url,
     deposit_request: &PayloadBody,
 ) -> Result<(), KairosClientError> {
+    let client = blocking::Client::new();
     let url = base_url.join("/api/v1/deposit").unwrap();
     let response = client
         .post(url)
         .header("Content-Type", "application/json")
         .json(deposit_request)
         .send()
-        .await
         .map_err(Into::<KairosClientError>::into)?;
     let status = response.status();
     if !status.is_success() {
--- a/kairos-cli/src/client.rs
+++ b/kairos-cli/src/client.rs
@@ -50,7 +50,7 @@ pub fn submit_transaction_request(
         .map_err(Into::<KairosClientError>::into)?;
     let status = response.status();
     if !status.is_success() {
-        Err(KairosClientError::ResponseError(status.to_string()))
+        Err(KairosClientError::KairosServerError(status.to_string()))
     } else {
         Ok(())
     }
--- a/kairos-cli/src/client.rs
+++ b/kairos-cli/src/client.rs
@@ -1,0 +1,58 @@
+use kairos_server::routes::PayloadBody;
+
+use reqwest::{Client, Url};
+use serde::{Deserialize, Serialize};
+use std::fmt;
+
+#[derive(PartialOrd, Ord, PartialEq, Eq, Debug, Serialize, Deserialize)]
+pub enum KairosClientError {
+    ResponseError(String),
+    ResponseErrorWithCode(u16, String),
+    DecodeError(String),
+    KairosServerError(String),
+}
+
+impl std::error::Error for KairosClientError {}
+
+impl fmt::Display for KairosClientError {
+    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let json_string = serde_json::to_string(self).map_err(|_| fmt::Error)?;
+        write!(formatter, "{}", json_string)
+    }
+}
+impl From<reqwest::Error> for KairosClientError {
+    fn from(error: reqwest::Error) -> Self {
+        let error_without_url = error.without_url();
+        if error_without_url.is_decode() {
+            KairosClientError::DecodeError(error_without_url.to_string())
+        } else {
+            match error_without_url.status() {
+                Option::None => Self::ResponseError(error_without_url.to_string()),
+                Option::Some(status_code) => {
+                    Self::ResponseErrorWithCode(status_code.as_u16(), error_without_url.to_string())
+                }
+            }
+        }
+    }
+}
+
+pub async fn submit_transaction_request(
+    client: &Client,
+    base_url: &Url,
+    deposit_request: &PayloadBody,
+) -> Result<(), KairosClientError> {
+    let url = base_url.join("/api/v1/deposit").unwrap();
+    let response = client
+        .post(url)
+        .header("Content-Type", "application/json")
+        .json(deposit_request)
+        .send()
+        .await
+        .map_err(Into::<KairosClientError>::into)?;
+    let status = response.status();
+    if !status.is_success() {
+        Err(KairosClientError::ResponseError(status.to_string()))
+    } else {
+        Ok(())
+    }
+}
--- a/kairos-cli/src/commands/deposit.rs
+++ b/kairos-cli/src/commands/deposit.rs
@@ -1,9 +1,14 @@
+use crate::client;
 use crate::common::args::{AmountArg, PrivateKeyPathArg};
 use crate::error::CliError;
+
+use reqwest::Url;
 
 use kairos_crypto::error::CryptoError;
 use kairos_crypto::implementations::Signer;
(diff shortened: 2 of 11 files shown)


kairos-cli/src/lib.rs
```diff
@@ -1,4 +1,44 @@
...
+}
+
+pub fn run(
+    Cli {
+        command,
+        kairos_server_address,
+    }: Cli,
+) -> Result<String, CliError> {
+    let kairos_server_address =
+        kairos_server_address.unwrap_or(Url::parse("http://0.0.0.0:9999").unwrap());
```
Avi-D-coder, line 38: `http://0.0.0.0:9999`?

marijanp: I thought I'll make the `kairos-server-address` an optional argument, and if users don't specify it, we default to some address. I chose a random local one. I'm open to making this a mandatory argument and then users would always need to specify a URL

kairos-cli/src/client.rs
```diff
@@ -0,0 +1,58 @@
...
+        .json(deposit_request)
+        .send()
+        .await
+        .map_err(Into::<KairosClientError>::into)?;
+    let status = response.status();
+    if !status.is_success() {
+        Err(KairosClientError::ResponseError(status.to_string()))
+    } else {
+        Ok(())
+    }
```
koxu1996, line 57: Consider simplified flow:

```suggestion
    if !status.is_success() {
        return Err(KairosClientError::ResponseError(status.to_string()))
    }

    Ok(())
```

kairos-cli/src/client.rs
```diff
@@ -0,0 +1,58 @@
...
+use reqwest::{Client, Url};
+use serde::{Deserialize, Serialize};
+use std::fmt;
+
+#[derive(PartialOrd, Ord, PartialEq, Eq, Debug, Serialize, Deserialize)]
+pub enum KairosClientError {
+    ResponseError(String),
+    ResponseErrorWithCode(u16, String),
+    DecodeError(String),
+    KairosServerError(String),
```
koxu1996, line 12: Please remove/comment unused variant.

Rom3dius:
I'm a little confused as to why you're spawning a tokio runtime to use reqwests async API in a CLI application when reqwest has a blocking API and the nature of a CLI is generally synchronous.

Let's merge it, just leaving this comment here for future reference and to make everyone aware of reqwests blocking API.

kairos-cli/src/commands/deposit.rs
```diff
@@ -15,14 +20,31 @@ pub struct Args {
...
-    // TODO: Send transaction to the network, using Rust SDK.
+    let payload = SigningPayload::new_deposit(amount)
+        .try_into()
+        .expect("Failed serialize the deposit payload to bytes");
+    let signature = signer.sign(&payload)?;
+    let deposit_request = PayloadBody {
+        public_key,
+        payload,
+        signature,
+    };
```
koxu1996, line 39: I think it would be better to move `PayloadBody` construction into `submit_transaction_request()` at `kairos-cli/src/client.rs`. CLI subcommand - deposit in this case - should not be aware of REST details, client should.

kairos-cli/src/commands/deposit.rs
```diff
@@ -15,14 +20,31 @@ pub struct Args {
...
-pub fn run(args: Args) -> Result<String, CliError> {
-    let _amount: u64 = args.amount.field;
-    let _signer =
+pub fn run(args: Args, kairos_server_address: Url) -> Result<String, CliError> {
+    let amount: u64 = args.amount.field;
+    let signer =
         Signer::from_private_key_file(args.private_key_path.field).map_err(CryptoError::from)?;
 
-    // TODO: Create transaction and sign it with `signer`.
+    let client = reqwest::Client::new();
```
koxu1996, line 28: Similarily to https://github.com/cspr-rad/kairos/pull/91/files#r1590918696, I would push `reqwest` client construction down to client implementation at `kairos-cli/src/client.rs`.

kairos-cli/src/lib.rs
```diff
@@ -1,4 +1,44 @@
...
+use clap::{Parser, Subcommand};
+use reqwest::Url;
+
+#[derive(Parser)]
+#[command(name = "Kairos Client", about = "CLI for interacting with Kairos")]
+pub struct Cli {
+    #[command(subcommand)]
+    pub command: Command,
+    #[arg(long, value_name = "URL")]
+    pub kairos_server_address: Option<Url>,
```
koxu1996, line 18: We always need server address, thus it should not be an `Option`.

**[Request]** Use the default value, for example:

```suggestion
    #[arg(long, value_name = "URL", default_value_t = get_default_api_address())]
    pub kairos_server_address: Url,
```

It would simplify `kairos_cli::run` - there is no longer need to destruct `Cli` structure.

kairos-cli/src/lib.rs
koxu1996: It should be mandatory argument, but with default value, so user can skip it - see https://github.com/cspr-rad/kairos/pull/91/files#r1590933505.

koxu1996 (changes requested):
Needs more work - details in comments.

koxu1996:
> I'm a little confused as to why you're spawning a tokio runtime to use reqwests async API in a CLI application when reqwest has a blocking API and the nature of a CLI is generally synchronous.
> 
> Let's merge it, just leaving this comment here for future reference and to make everyone aware of reqwests blocking API.

@Rom3dius that's right! **CLI client** (replaced with Rust SDK in the future) should be synchronous, not only for simplicity, but there is no advantage of having async there. @marijanp could you use `reqwest::blocking`?

marijanp:
@Rom3dius @koxu1996 I began writing the `client` with potentially factoring it out in a separate crate, so I decided to make it async as they usually  don't block but do so in our CLI. But it might just be premature optimization. so I will make this blocking.

kairos-cli/src/client.rs
Avi-D-coder: I prefer having the else clause.

Avi-D-coder (approved):

kairos-cli/src/commands/deposit.rs
marijanp: The api of `submit_transaction_request` should in future depending on the type of transaction we are doing ideally change the endpoint to either `deposit`, `transfer` and `withdraw` and could be used for all of these types.
If I make the PayloadBody construction part of the `submit_transaction_request` the api would be
```Rust
submit_transaction_request(public_key: Vec<u8>, payload: Vec<u8>, signature: Vec<u8>)
```
which could accidentally be invoked like:
```Rust
submit_transaction_request(payload, signature, public_key)
```
Which is not as safe as making the `PayloadBody` construction part of the function I would argue.

kairos-cli/src/commands/deposit.rs
marijanp: [31b4c71](https://github.com/cspr-rad/kairos/pull/91/commits/31b4c715ca55533de1d583e8ad434c55f397b78c)

kairos-cli/src/client.rs
marijanp: It should actually be used when the server returns an error, so it's unused by accident. It's used now here [9d0ae91](https://github.com/cspr-rad/kairos/pull/91/commits/9d0ae91ca85582c932c3a81a8c527628baaf8392)

kairos-cli/src/lib.rs
marijanp: Using `default_value` now [987177d](https://github.com/cspr-rad/kairos/pull/91/commits/987177d0020214b76a8d3f346e516245a12f5e86)

Avi-D-coder (approved):

kairos-cli/src/lib.rs
koxu1996: Code removed in https://github.com/cspr-rad/kairos/pull/91/commits/0c980efa500346d61c26d90b31d252e9e8403b5e.

kairos-cli/src/commands/deposit.rs
koxu1996: Not really. In the future we should have single endpoint for handling all types of transactions - https://github.com/cspr-rad/kairos/issues/97. Also, once full transaction format is adopted in server, then it would become `submit_transaction_request(full_tx)`.

However, when #23 is done there will be no REST client in `kairos-cli` at all. So I consider this is only temporary code, no reason to discuss it longer.

koxu1996 (approved):

PR #93, opened 2024-05-06
clean shutdown + default log level INFO

Rom3dius:
Two changes:

1. Implements CTRL+C and kill signal handler to kairos-server, which axum then uses to stop serving new connections and finish serving existing connections before shutting down.

2. Sets the default logging level to INFO, otherwise you can't see any logs unless you use the RUST_LOG environment variable.


koxu1996 (approved):
LGTM

Avi-D-coder:
If your using vscode please enable `.editorconfig`. It keeps the line endings consistent.
https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig

Avi-D-coder (approved):

PR #94, opened 2024-05-07
deposit: refactor to forward deploys to L1

marijanp:
## Motivation
- Checking signatures and adding the deposit to the batch is not required to be done in the deposit endpoint.
- It should rather according to casper best-practices forward deploys to the L1.
- This deposit-deploy will be prepared in the kairos-cli and signed
- The signed deploy is sent to the L1 through kairos
- Once the L1 processes the deploy, an event will be triggered.
- The kairos server will then add the deposit to the batch.

## TODOS

- 1. [x] Don't break the other server tests, i.e. implement alternative `deposit` handler which can be used for property tests and only adds deposits to the batch
- 2. [x] Integrate the deposit contract i.e. wait for it to be merged.
- 3. [x] Implement the `kairos-cli` counterpart preparing the deposit deploy
- 4. [x] Update the end-to-end-test

## Successive PR
- [ ] use connect casper-event-toolkit with kairos-server, add deposit to batch see #110


jonas089:
"Checking signatures and adding the deposit to the batch is not required to be done in the deposit endpoint."

Was this ever up for debate? 

The Kairos server/ node will query the L1 for deposits and then roll them up in a batch together with Transactions and Withdrawals and produce a proof.

The proof is then submitted to another entry point of the demo smart contract that will verify it, extract the new root from the journal and update the contract state to store that new root.

jonas089:
"It should rather according to casper best-practices forward deploys to the L1."

But native L1 deposits will still occur, so not every deposit will be forwarded from the L2 to the L1? 

If you want to route Deposits through the L2 then yes this is the way to do it, but that's an additional means of depositing and not a replacement?

marijanp:
@jonas089 the current implementation of the server suggests something different https://github.com/cspr-rad/kairos/blob/27ef3fc980eef4e099a06f192cd68c14729ba3a9/kairos-server/src/routes/deposit.rs#L24-L52

jonas089:
> @jonas089 the current implementation of the server suggests something different https://github.com/cspr-rad/kairos/blob/27ef3fc980eef4e099a06f192cd68c14729ba3a9/kairos-server/src/routes/deposit.rs#L24-L52

But if you forward a signed deploy to the L2 you might as well use the CLI to forward the deploy directly to the L1.

It will be recorded as an Event and then picked up by the L2.

Handling it differently introduces unwanted complexity.

marijanp:
@jonas089 If you ask me I would get rid of the `deposit` endpoint entirely. However [the casper docs](https://docs.casper.network/developers/dapps/technology-stack/#signing-transactions) suggests that we should forward through a backend.
Moreover @Avi-D-coder is interested in having an endpoint to be able to do low-setup property-tests.

jonas089:
> @jonas089 If you ask me I would get rid of the `deposit` endpoint entirely. However [the casper docs](https://docs.casper.network/developers/dapps/technology-stack/#signing-transactions) suggests that we should forward through a backend.
> Moreover @Avi-D-coder is interested in having an endpoint to be able to do low-setup property-tests.

Okay, shouldn't make a big difference, it'll work either way.

Avi-D-coder (approved):

jonas089 (approved):
lgtm

koxu1996:
I would avoid compile-time environment variables, that breaks compilation outside of Nix. It's also annoying to see analyzer errors in IDE: 

image

marijanp:
So if you don't have this environment variable sourced in your development environment and run the code it will instead throw a runtime exception. Also sourcing environment variables is not accomplished with Nix but rather captured with Nix. My local development environment sources these environment variables for me automatically when I enter the kairos repository.

PR #96, opened 2024-05-09
Periodic L1 synchronization

koxu1996:
This PR adds periodic L1 synchronization :arrows_clockwise:, based on [casper-event-toolkit](https://github.com/koxu1996/casper-event-toolkit).

### Details

The core component is `EventManager`, which allows tracking events from a specified smart contract. The process begins with the setup of CES-related data - schemas, toolkit fetcher, and the number of events - using `initialize()`. Subsequently, `process_new_events()` checks for new events based on its internal counter, `next_event_id`.

`L1SyncService` operates the event manager as a Tokio task and provides a safe API for shared access within an Axum server.

### Future work

- Once the deposit contract is merged into `main`, we will be able to parse deposit data from events - currently, this data is mocked - and push it to the batch manager (trie).
- Right now synchronization is called periodically - it will be replaced with SSE trigger in the future.

## Demo :movie_camera:

https://github.com/cspr-rad/kairos/assets/6506780/d7793f3c-6a98-41ec-8cb7-f8204588c075

If you want to reproduce it locally, please see the next section.

### Test environment

I used dockerized NCTL with bunch of shell scripts to run this demo - all available in [this gist](https://gist.github.com/koxu1996/c53011eb69e22561aff76388b1190864).

#### 1. Cleanup old resources and compile example smart contract

```sh
$ ./cleanup.sh
$ ./prepare.sh
```

#### 2. Run NCTL and deploy WASM

```sh
$ ./setup.sh
> Running NCTL container.
9e4fd0c153e63765ea13b565c70e2d2986ad929a2e74e74cb5e37938e3fe1edf

> Copying user keys.
Successfully copied 37.4kB to /tmp/casper-users

> Deploying session code.
- hash: f9531e06d8f533aa033106b40030ff6d4a9f6b63dc542ff3177d6d29412313ee

> Waiting for execution.
- attempt: 1/10
- attempt: 2/10
> Getting state root hash.
 - d196121dc9613867e228fb807c798fee4e9d1d7d84a2ba6e1a9067494a63c231

> Getting account hash.
 - c6ec39ce20e3bda02cd061e5c91698b0be8a5d5f9abf2b95b8a2a639ae88d060

> Getting contract hash.
 - 9397c7dbfe9ffd60fd03774d8b570bcb3a3a482c5eb3202c97779cb5c0dac008
```

At this point you have **contract hash** that is ready to be used in the next steps.

#### 3. Initialize Casper Event Standard

```sh
$ ./call-init-entrypoint.sh 9397c7dbfe9ffd60fd03774d8b570bcb3a3a482c5eb3202c97779cb5c0dac008
> Calling 'init' entrypoint.
 - 24c05e37de1c453dd85ef9d93384583c3bbc7106dc916b219e02335cfd3a78ff

> Waiting for execution.
- attempt: 1/20
- attempt: 2/20
- attempt: 3/20
- attempt: 4/20

*-------*
| DONE! |
*-------*
```

#### 4. Call deposit/batch entrypoint to emit test events

```sh
./call-batch-entrypoint.sh 9397c7dbfe9ffd60fd03774d8b570bcb3a3a482c5eb3202c97779cb5c0dac008
> Calling 'batch' entrypoint.
 - 63342404ed21eab1d3a29703a76f81834a0e8bb3bdf3ad1caa53357bf2bd4c38

> Waiting for execution.
- attempt: 1/20
- attempt: 2/20
- attempt: 3/20
- attempt: 4/20
- attempt: 5/20

*-------*
| DONE! |
*-------*
```

### Additional changes for demo

In order to record nice and quick demo, I updated subscribed to include relevant logs:

```patch
diff --git a/kairos-server/src/main.rs b/kairos-server/src/main.rs
index 107e421..7b8f1ce 100644
--- a/kairos-server/src/main.rs
+++ b/kairos-server/src/main.rs
@@ -5,6 +5,11 @@ use kairos_server::config::ServerConfig;
 async fn main() {
     let subscriber = tracing_subscriber::fmt()
         .with_max_level(tracing::Level::INFO)
+        .with_env_filter(
+            "kairos_server::l1_sync=debug,kairos_server=warn"
+                .parse::<tracing_subscriber::EnvFilter>()
+                .unwrap(),
+        )
         .finish();

     tracing::subscriber::set_global_default(subscriber).expect("Failed to set subscriber");
```

I also reduced synchronization interval from 30 to 10 seconds:

```patch
diff --git a/kairos-server/src/l1_sync/interval_trigger.rs b/kairos-server/src/l1_sync/interval_trigger.rs
index 419f13c..08acc65 100644
--- a/kairos-server/src/l1_sync/interval_trigger.rs
+++ b/kairos-server/src/l1_sync/interval_trigger.rs
@@ -5,7 +5,7 @@ use std::sync::Arc;
 use super::service::L1SyncService;

 pub async fn run(sync_service: Arc<L1SyncService>) {
-    let mut interval = time::interval(Duration::from_secs(30));
+    let mut interval = time::interval(Duration::from_secs(10));

     loop {
         interval.tick().await;
```


Rom3dius (approved):
solid work

kairos-server/src/l1_sync/service.rs
```diff
@@ -0,0 +1,102 @@
...
+    let mut em = event_manager.lock().await;
+
+    match command {
+        SyncCommand::Initialize(rpc_url, contract_hash, completion_ack) => {
+            em.initialize(&rpc_url, &contract_hash).await?;
+            let _ = completion_ack.send(());
+        }
+        SyncCommand::TriggerSync(completion_ack) => {
+            em.process_new_events().await?;
+            let _ = completion_ack.send(());
```
Avi-D-coder, line 97: what happens on failure?

Avi-D-coder (changes requested):

kairos-server/src/l1_sync/interval_trigger.rs
```diff
@@ -0,0 +1,20 @@
...
+
+    loop {
+        interval.tick().await;
+
+        tracing::debug!("Triggering periodic L1 sync");
+        let result = sync_service.trigger_sync().await;
+
+        if let Err(e) = result {
+            tracing::error!("Unable to trigger sync: {}", e);
+        }
```
marijanp, line 18: Nit: use `map_err`

marijanp (changes requested):

kairos-server/src/l1_sync/service.rs
```diff
@@ -0,0 +1,102 @@
...
+    event_manager: Arc<Mutex<EventManager>>,
+) {
+    tracing::debug!("Event manager running and waiting for commands");
+    while let Some(command) = rx.recv().await {
+        match handle_command(command, event_manager.clone()).await {
+            Ok(()) => {}
+            Err(L1SyncError::UnexpectedError(e)) => panic!("Unrecoverable error: {}", e),
+            Err(e) => tracing::error!("Transient error: {}", e),
+        }
+    }
```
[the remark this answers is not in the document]
koxu1996: Done in https://github.com/cspr-rad/kairos/pull/96/commits/1fac7159cb672b505afb2b1fa3d8024855cf29f5.

kairos-server/src/l1_sync/interval_trigger.rs
koxu1996: Done in https://github.com/cspr-rad/kairos/pull/96/commits/be86df8504c256394e59ca2ee20cca25424ee819.

kairos-server/src/lib.rs
```diff
@@ -26,8 +28,37 @@ pub fn app_router(state: Arc<state::BatchStateManager>) -> Router {
...
+
+    // Run layer 1 synchronization.
+    // TODO: Replace interval with SSE trigger.
+    let l1_sync_service = Arc::new(L1SyncService::new(batch_service).await);
+    tokio::spawn(async move {
+        if let Err(e) = l1_sync_service
+            .initialize(config.casper_rpc.to_string(), config.casper_contract_hash)
+            .await
+        {
+            panic!("Event manager failed to initialize: {}", e);
```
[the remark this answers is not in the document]
koxu1996: `map_err` introduced in https://github.com/cspr-rad/kairos/pull/96/commits/326b147f592f522fc311b56140fd860fff4d5307.

kairos-server/src/lib.rs
```diff
@@ -26,8 +28,37 @@ pub fn app_router(state: Arc<state::BatchStateManager>) -> Router {
...
+pub async fn run_l1_sync(config: ServerConfig, batch_service: Arc<BatchStateManager>) {
+    // Make sure real contract hash was provided.
+    if config.casper_contract_hash
+        == "0000000000000000000000000000000000000000000000000000000000000000"
+    {
+        tracing::warn!(
+            "Casper contract hash not configured, L1 synchronization will NOT be enabled."
+        );
+        return;
+    }
```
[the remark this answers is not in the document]
koxu1996: I am expecting that contract hash is always required - same as the RPC url. Is there any reason to run Kairos server without observing L1 state? I think no, so I introduced default value `00...00` with warning if unchanged.

Another reason to avoid `Option` for config field, is that it requires creating newtype with  `FromStr` implementation. IMO code gets unnecessarily complicated, you can see changes preview here - https://github.com/cspr-rad/kairos/commit/74adecdd9959aa0dc5364af3ef0abe8ed8a7c1ff. I would prefer to avoid it, but let me know if you like it better.

kairos-server/src/l1_sync/event_manager.rs
```diff
@@ -0,0 +1,85 @@
...
+            ces_metadata: metadata,
+        };
+        let schemas = fetcher.fetch_schema().await?;
+        tracing::debug!("Schemas fetched successfully");
+
+        self.fetcher = Some(fetcher);
+        self.schemas = Some(schemas);
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
koxu1996: > The logic from `initialize` can go into `new` the mutex.

I don't want initialization logic in the constructor, as it should be straightforward and non-blocking.

> The `EventManager` service should start initialized, it should not be a command.

I initially made it a command, because I wanted to run initialization in the background, without blocking server start. After thinking it out again I see that we should synchronize state, before processing requests.

Changes done in https://github.com/cspr-rad/kairos/pull/96/commits/1b6cae2cd5610e853be8ed1439568ddcfb7eaa95.

kairos-server/src/l1_sync/service.rs
koxu1996: Fixed in https://github.com/cspr-rad/kairos/pull/96/commits/f437b2e16ab9d3bfd194a113c11a4b8b059a955d.

koxu1996:
@Avi-D-coder @marijanp Ready for re-review.

kairos-server/src/l1_sync/event_manager.rs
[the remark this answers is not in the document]
Avi-D-coder: I'd still like to see `initialize` merged into `new`.
But doing that in an eventual followup is fine, so I'll approve.

Avi-D-coder (approved):

kairos-server/src/l1_sync/event_manager.rs
```diff
@@ -0,0 +1,89 @@
...
+
+impl EventManager {
+    pub fn new(server_state: Arc<ServerStateInner>) -> Self {
+        EventManager {
+            next_event_id: 0,
+            fetcher: None,
+            schemas: None,
+            server_state,
+        }
+    }
```
marijanp, line 27: I don't see where it makes sense to model the `fetcher` and `schemas` attributes as optionals and introduce a new type that basically sets up the user for using the api in a wrokng way, and moreover forces a user to have a two-step setup process ending up with these two possible use scenarios:
(Notice that the case `fetcher == Some && schema == None` and `fetcher == None && schema Some` is not even possible)
```rust
// correct
let event_manager = EventManager::new(); // fetcher == None && schema == None
event_manager.initialize(); // fetcher == Some(..) && schema == Some(..)
event_manager.process_new_events();

// wrong usage, but possible without any compiler warning
let event_manager = EventManager::new();  // fetcher == None && schema == None
event_manager.process_new_events(); // This will fail and I think I should not even be able to call this and a strong indicator is a missing fetcher and schema, which can both be used to prevent an eventuall call

// better: process_new_events is now just a function in the server crate and can belong semantically to the server
// the ServerState contains all the information like the `rpc_url`, `contract_hash`, `server_state`, even the `next_event_id`
pub fn process_new_events(server_state: &ServerState, callback: FnOnce -> Result) -> Result {
// initialize the fethcer. Could fail, hence result return type
// initialize the schema. Could fail, hence result return type
// call callback which injects any behavior a user might want to use which could potentially fail, hence Result return type of the callback
}
```
The latter being a single function call, which deals with all the details a user does not care about. The user only cares about what should happen when new events are observed, reducing the amount of knowledge to just implement the callback logic.

marijanp (changes requested):

kairos-server/src/l1_sync/event_manager.rs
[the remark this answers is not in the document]
koxu1996: Fixed in https://github.com/cspr-rad/kairos/pull/96/commits/50a3654c33e72267c89df53df61ca72a2bcca9eb.

kairos-server/src/lib.rs
```diff
@@ -26,8 +28,37 @@ pub fn app_router(state: Arc<state::BatchStateManager>) -> Router {
...
+    // TODO: Replace interval with SSE trigger.
+    let l1_sync_service = Arc::new(L1SyncService::new(batch_service).await);
+    tokio::spawn(async move {
+        if let Err(e) = l1_sync_service
+            .initialize(config.casper_rpc.to_string(), config.casper_contract_hash)
+            .await
+        {
+            panic!("Event manager failed to initialize: {}", e);
+        }
+        l1_sync::interval_trigger::run(l1_sync_service).await;
```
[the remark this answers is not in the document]
koxu1996: Initialization steps removed in https://github.com/cspr-rad/kairos/pull/96/commits/50a3654c33e72267c89df53df61ca72a2bcca9eb and https://github.com/cspr-rad/kairos/pull/96/commits/0d4fb1394388a89fb8701246b34f8eb067a4268e.

kairos-server/src/l1_sync/event_manager.rs
koxu1996: I removed partial construction pattern, so `fetcher` and `schemas` are no longer optional: https://github.com/cspr-rad/kairos/pull/96/commits/50a3654c33e72267c89df53df61ca72a2bcca9eb.

kairos-server/src/l1_sync/event_manager.rs
```diff
@@ -0,0 +1,73 @@
...
+            let txn = KairosTransaction::Deposit(L1Deposit {
+                amount: 100,
+                recipient,
+            });
+            self.server_state
+                .batch_state_manager
+                .enqueue_transaction(txn)
+                .await
+                .map_err(|e| L1SyncError::UnexpectedError(format!("unable to batch tx: {}", e)))?;
+            self.next_event_id = i + 1;
```
marijanp, line 68: This reads odd to me, I would move this outside of the for loop, and set it to `fetch_events_count`. We are not benefiting from updating that value immediately.

kairos-server/src/l1_sync/event_manager.rs
```diff
@@ -0,0 +1,73 @@
...
+
+        let num_events = self.fetcher.fetch_events_count().await?;
+        for i in self.next_event_id..num_events {
+            let event = self.fetcher.fetch_event(i, &self.schemas).await?;
+            tracing::debug!("Event {} fetched: {:?}.", i, event);
+
+            // TODO: Parse full transaction data from event, then push it to Data Availability layer.
+
+            // TODO: Once we have ASN transaction, it should be converted and pushed into batch.
+            let recipient: Vec<u8> = "cafebabe".into();
```
Avi-D-coder, line 58: why are we mocking the deposit here?

Avi-D-coder (changes requested):
please fix the mocked recipient.

kairos-server/src/l1_sync/event_manager.rs
```diff
@@ -0,0 +1,73 @@
...
+    next_event_id: u32,
+    fetcher: Fetcher,
+    schemas: Schemas,
+    server_state: Arc<ServerStateInner>,
+}
+
+impl EventManager {
+    pub async fn new(
+        rpc_url: &str,
+        contract_hash: &str,
```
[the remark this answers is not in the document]
koxu1996: Fixed in https://github.com/cspr-rad/kairos/pull/96/commits/736a540d4b23c957162231f2d267d7f7a050ad2c and https://github.com/cspr-rad/kairos/pull/96/commits/3b52fb1ff889beed3c00504f519f649218bee8d4.

kairos-server/src/l1_sync/service.rs
```diff
@@ -0,0 +1,82 @@
...
+
+pub struct L1SyncService {
+    command_sender: mpsc::Sender<SyncCommand>,
+    //event_manager_handle: tokio::task::JoinHandle<()>,
+}
+
+impl L1SyncService {
+    pub async fn new(
+        rpc_url: String,
+        contract_hash: String,
```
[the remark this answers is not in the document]
koxu1996: Fixed in https://github.com/cspr-rad/kairos/pull/96/commits/736a540d4b23c957162231f2d267d7f7a050ad2c and https://github.com/cspr-rad/kairos/pull/96/commits/3b52fb1ff889beed3c00504f519f649218bee8d4.

kairos-server/src/lib.rs
```diff
@@ -38,6 +40,35 @@ pub fn app_router(state: ServerState) -> Router {
         .with_state(state)
 }
 
+pub async fn run_l1_sync(config: ServerConfig, server_state: Arc<ServerStateInner>) {
```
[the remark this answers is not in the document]
koxu1996: `config` removed in https://github.com/cspr-rad/kairos/pull/96/commits/8c5cc97791663835682a8b1d1e89dac1c730278e.

kairos-server/src/l1_sync/event_manager.rs
koxu1996: I want to keep `EventManager` state - `next_event_id` - in sync with enqueued transactions. In other words if some transaction fails to batch, then manager does not consider event as processed.

kairos-server/src/l1_sync/event_manager.rs
koxu1996: I was hoping that https://github.com/cspr-rad/kairos/pull/88 will get merged, and then it could be used for parsing. Since `kairos-tx` in the contract is postponed for later discussion, I will implement basic parsing here.

koxu1996: #121 got merged, so I introduced `Deposit` event parsing in https://github.com/cspr-rad/kairos/pull/96/commits/ff7b6c69621e284696c10e460d8f35cdce9ae2b3.

However, `recipient` field is still mocked. It is not currently possible to obtain depositor **public key** from event, as it contains **account hash** (unnecessarily wrapped in `Key`)... This was already solved in [`kairos-tx` for contract PR](https://github.com/cspr-rad/kairos/pull/88/commits/75b5143602fa2bf3498044489b46e9d3277169c0#diff-a69ffc4ce0a99c02c5f8feb4f57529efd0c1871a711bad036378a977c1d9b2bfR55-R63) - validation if explicitly given public key matches with caller account hash - so maybe we should reconsider merging it before demo?

PR #98, opened 2024-05-13
Risc0 circuit

Avi-D-coder:
This is the risc0 circuit that me and @jonas089 worked on, the branch is based on @marijanp's nixified risc0 toolchain.

### TODO (not necessarily before merge):
- groth16 wrapping from jonas089's branch.
- ~tests.~
- Switch to bincode2.0rc3 from json.
- Merge the logic in the server with the logic in `kairos-circuit-logic`.
  The logic in the server does a bunch of prechecks that are important in that setting.
- get asn1 compiling in risc0-guest for signature verification.
- ~use patched sha256 crate from risc0~


jonas089:
I personally liked my initial structure a lot better.
It was much easier to work with and easier to read.
kairos-prover-server is an unnecessary crate and all of this should happen inside the kairos-server.

It also doesn't look like there are any tests for the proving process.
If I just overlooked them please point them out.

Avi-D-coder:
No tests yet.
kairos-prover-server is needed, it could be a CLI instead as marijanp has suggested.

But they compile with different toolchains so some separation is required.
I made it a microservice because I expect proving would occur on machines with GPUs.

The bare minimum logic is in the risc0 workspace since that workspace is still broken for RA.

jonas089:
> No tests yet. kairos-prover-server is needed, it could be a CLI instead as marijanp has suggested.
> 
> But they compile with different toolchains so some separation is required. I made it a microservice because I expect proving would occur on machines with GPUs.
> 
> The bare minimum logic is in the risc0 workspace since that workspace is still broken for RA.

Also how am I supposed to construct a Snapshot<Account> when there are not tests?

jonas089:
Also how am I supposed to construct a Snapshot when there are not tests?

jonas089:
> No tests yet. kairos-prover-server is needed, it could be a CLI instead as marijanp has suggested.
> 
> But they compile with different toolchains so some separation is required. I made it a microservice because I expect proving would occur on machines with GPUs.
> 
> The bare minimum logic is in the risc0 workspace since that workspace is still broken for RA.

Proving a Batch should not be an API endpoint. It should happen automatically on the kairos-server every n seconds / era / transactions.

jonas089 (changes requested):
Without tests this can't be merged. We need an integration test for a scenario where a few Transactions have occurred and the batch proof is generated. Then the proof should be verified.

Such a tests is a requirement to 
1. see that the circuit code works
2. develop the contract entry point that verifies the proof

Avi-D-coder:
> Also how am I supposed to construct a Snapshot when there are not tests?

fine grained tests will come soon, snapshot is constructed by existing code in the server (batch manager) if you want to take a crack at constructing one take a look at that, and or the property tests i have in the trie. We need a good property test generator like i have in the ops trie tests.

with that being said if you look at the signature of the trait I introduced in the prover server and implement in the prover, that is what you should build the contract end point around.

it should provide snapshot, ProofInputs and handle the output.

jonas089:
> > Also how am I supposed to construct a Snapshot when there are not tests?
> 
> fine grained tests will come soon, snapshot is constructed by existing code in the server (batch manager) if you want to take a crack at constructing one take a look at that, and or the property tests i have in the trie. We need a good property test generator like i have in the ops trie tests.
> 
> with that being said if you look at the signature of the trait I introduced in the prover server and implement in the prover, that is what you should build the contract end point around.
> 
> it should provide snapshot, ProofInputs and handle the output.

I can't begin to write the contract entry point before this PR has been merged and this PR can't be merged without tests. The entry point will depend on types exposed by this PR so therefore we need to merge it first and it must be obvious how a proof is constructed and we must know for sure that the circuit code works as is.

Avi-D-coder:
i'd suggest pulling this branch into a branch based on the refactored deposit contract and following the types from there.

alternatively just copy the ProofInputs/Outputs types directly into your file and don't worry about the function that turns one to the other. just make the body todo and make the logic all type check and lint.

I can write an initial prop test later today or tomorrow.

but don't let that block you the proof Inputs and Outputs types are fixed. if you write the logic around those types it will work.

jonas089:
> i'd suggest pulling this branch into a branch based on the refactored deposit contract and following the types from there.
> 
> alternatively just copy the ProofInputs/Outputs types directly into your file and don't worry about the function that turns one to the other. just make the body todo and make the logic all type check and lint.
> 
> I can write an initial prop test later today or tomorrow.
> 
> but don't let that block you the proof Inputs and Outputs types are fixed. if you write the logic around those types it will work.



> i'd suggest pulling this branch into a branch based on the refactored deposit contract and following the types from there.
> 
> alternatively just copy the ProofInputs/Outputs types directly into your file and don't worry about the function that turns one to the other. just make the body todo and make the logic all type check and lint.
> 
> I can write an initial prop test later today or tomorrow.
> 
> but don't let that block you the proof Inputs and Outputs types are fixed. if you write the logic around those types it will work.

This workflow is incredibly messy, we already have too many branches and I'm sure that most of us have lost track of what's happening on them.

Avi-D-coder:
> This workflow is incredibly messy

it is indeed, maybe one day pijul or stacked diffs will save us, but until then we have to live with the trade offs. complete features in a single pr are going to result in branches that live a while. on the other hand merging incomplete code results in churn and quality issues.

copying two types is about the minimum amount of coupling between branches you could do. but yes it's always a mess, until the rebase.

marijanp (commented):
First part of the review, I will need some help understanding the AccountTrie and the Transaction types

cargo-update-workspaces.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env sh
+
+# Set the path to the first argument, defaulting to the current directory if no argument is provided
+path="${1:-.}"
+
+# Find and update Cargo.lock files starting from the specified path or the default path
+find "$path" -name Cargo.lock -execdir cargo update --verbose \;
```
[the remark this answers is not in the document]
Avi-D-coder: Yes, find has poor syntax.

kairos-prover/kairos-tx/Cargo.toml
```diff
@@ -7,9 +7,9 @@ license.workspace = true
 [lib]
 
 [dependencies]
-num-traits = "0.2"
-rasn = { version = "0.12", default-features = false, features = ["macros"] }
-sha2 = "0.10"
+num-traits = { version = "0.2", default-features = false }
+rasn = { version = "0.14", default-features = false, features = ["macros"] }
+sha2 = { version = "0.10", default-features = false }
```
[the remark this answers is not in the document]
Avi-D-coder: If it's in the workspace we should not need to.

kairos-prover/kairos-circuit-logic/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+pub struct ProofOutputs {
+    pub pre_batch_trie_root: TrieRoot<NodeHash>,
+    pub post_batch_trie_root: TrieRoot<NodeHash>,
+    /// TODO consider replacing with a count and hash of the processed deposits
+    pub deposits: Box<[L1Deposit]>,
+    pub withdrawals: Box<[Signed<Withdraw>]>,
+}
+
+impl ProofInputs {
+    pub fn run_batch_proof_logic(self) -> Result<ProofOutputs, String> {
```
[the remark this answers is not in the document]
Avi-D-coder: It consumes self, so I made it associated.
It can also be found by method auto complete this way, if you have `ProofInputs`.

kairos-prover/kairos-prover-risc0-server/src/main.rs
```diff
@@ -0,0 +1,200 @@
...
+    tracing::info!("listening on `{}`", socket_addr);
+    axum::serve(listener, app).await.unwrap()
+}
+
+#[derive(TypedPath, Debug, Clone, Copy)]
+#[typed_path("/api/v1/prove/batch")]
+pub struct ProveBatch;
+
+pub async fn prove_batch_route(
+    _: ProveBatch,
```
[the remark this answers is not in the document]
Avi-D-coder:
```
#[derive(TypedPath, Debug, Clone, Copy)]
#[typed_path("/api/v1/prove/batch")]
pub struct ProveBatch;
```

This is the typed path, since the endpoint takes JSON, the there is no data from the URL, hence an empty struct.

kairos-prover/kairos-prover-risc0-server/src/main.rs
```diff
@@ -0,0 +1,200 @@
...
+            batches,
+            |proof_inputs| {
+                Ok(crate::prove_execution(proof_inputs)
+                    .map(|(proof_outputs, _)| proof_outputs)
+                    .expect("Failed to prove execution"))
+            },
+        );
+    }
+
+    #[test_strategy::proptest(ProptestConfig::default(), cases = 2)]
```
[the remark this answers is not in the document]
Avi-D-coder: Proving a single batch took 943.392243084s in GitHub darwin CI vs 6-12 seconds locally with metal.

Dev mode definitely dev mode.
I consider proving to be a test of risc0, while it would be nice to test our dependency, It's not practical in this case.

Using the risc0 dev mode we still test our logic, since risc0 still runs our logic and provides them in the receipt as proof outputs.

We can even increase the number of cases.

```
kairos-prover> 2024-05-30T20:45:04.997840Z  INFO kairos_prover_risc0_server: Proved batch: 943.392243084s
```

kairos-prover/kairos-circuit-logic/src/account_trie.rs
```diff
@@ -0,0 +1,376 @@
...
+            let ProofOutputs {
+                pre_batch_trie_root,
+                post_batch_trie_root,
+                deposits: _,
+                withdrawals: _,
+            } = proving_hook(proof_inputs).expect("Failed to prove execution");
+
+            assert_eq!(pre_batch_trie_root, prior_root_hash);
+            assert_eq!(post_batch_trie_root, new_root_hash);
+            prior_root_hash = new_root_hash;
```
marijanp, line 297: Can we return this instead of mutating it?

kairos-prover/kairos-circuit-logic/src/account_trie.rs
```diff
@@ -0,0 +1,376 @@
...
+        // Add the amount to the recipient's account
+        recipient_account.balance = recipient_account
+            .balance
+            .checked_add(transfer.amount)
+            .ok_or("recipient balance overflow")?;
+
+        Ok(())
+    }
+
+    pub fn deposit(&mut self, deposit: &L1Deposit) -> Result<(), TxnErr> {
```
marijanp, line 144: Why the choice to make this associated function take the `L1Deposit` struct directly but in `withdraw` and `transfer` you decompose the types. We could make it uniform and use the structs in `withdraw` and `transfer` too

kairos-prover/kairos-circuit-logic/src/account_trie.rs
```diff
@@ -0,0 +1,376 @@
...
+}
+
+impl PortableHash for Account {
+    fn portable_hash<H: PortableUpdate>(&self, hasher: &mut H) {
+        self.pubkey.portable_hash(hasher);
+        self.balance.portable_hash(hasher);
+    }
+}
+
+/// A utility function to hash multiple buffers reusing the same hasher.
```
marijanp, line 242: Maybe add a docstring that `out[i] = hash(item[i])`. I had to read the impl to understand it. First I thought that `out[i] = hash(item[1] <> hash(item[i-1]))`

kairos-prover/kairos-circuit-logic/src/transactions.rs
```diff
@@ -0,0 +1,541 @@
...
+
+    pub amount: u64,
+}
+
+/// TODO remove this struct
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Deposit {
+    pub amount: u64,
+}
```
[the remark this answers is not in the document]
Avi-D-coder: When I switch the server to use the circuit logic crate.

The server still uses the old deposit.

kairos-prover/kairos-circuit-logic/src/transactions.rs
```diff
@@ -0,0 +1,541 @@
...
+
+            writeln!(f, "AccountsState.l2.accounts: {:?}\n}}", self.l2.accounts)
+        }
+    }
+
+    impl Default for AccountsState {
+        fn default() -> Self {
+            unreachable!("AccountsState should always be created with AccountsState::new()");
+        }
+    }
```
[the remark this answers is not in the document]
Avi-D-coder: Never seen it before, but it successfully caught usage of proptest::any instead of any_with.

Proptest calls default if you don't provide a the argument.

I don't think I use AccountsState as a Arbitrary parameter any more but it's easy to accidentally do when writing proptests with the test-strategy macros.

kairos-prover/kairos-circuit-logic/src/transactions.rs
```diff
@@ -0,0 +1,541 @@
...
+                    (signed_withdraw(sender.to_vec()), TxnExpectedResult::Success)
+                }
+                _ => (signed_withdraw(sender.to_vec()), TxnExpectedResult::Failure),
+            }
+        }
+    }
+
+    #[derive(Debug, Clone)]
+    pub struct Accounts<A: Debug> {
+        pub pub_keys: Vec<Rc<PublicKey>>,
```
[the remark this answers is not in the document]
Avi-D-coder: we sample keys from the vector. hashmaps can't have an efficient random sample function. usize -> PublicKey

kairos-prover/kairos-circuit-logic/src/transactions.rs
```diff
@@ -0,0 +1,541 @@
...
+                (Some(l1_bal), Some(l2_bal)) => {
+                    *l1_balance = l1_bal;
+                    l2_account.balance = l2_bal;
+
+                    (L1Deposit { recipient, amount }, TxnExpectedResult::Success)
+                }
+                _ => {
+                    unreachable!("For now I am not testing the case where the deposit fails");
+                    // (L1Deposit { recipient, amount }, TxnExpectedResult::Failure)
+                }
```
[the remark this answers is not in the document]
Avi-D-coder: Testing for failures would expand the scope of the PR even more. I wanted to get the circuit logic merged to unblock other work before more rigorous testing, but ended up implementing most of the proptest random test model despite that. 

I had not intended this pr to have proptests at all. The proptest work was originally going to be follow-up because it's more complex than the actual circuit code. As it stands right now the test model matches only successful operations, failures would be different in a number of few known case.

kairos-prover/kairos-circuit-logic/src/account_trie.rs
Avi-D-coder: no we have to provide the new root to the next batch that run in the next iteration of the loop.

kairos-prover/kairos-circuit-logic/src/transactions.rs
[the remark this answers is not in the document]
Avi-D-coder: Removed whole impl, since I no longer use `AccountsState` as `Arbitrary::Parameters`.

kairos-prover/kairos-circuit-logic/src/account_trie.rs
Avi-D-coder: `L1Deposit` contains the recipient public key, and amount, it's only a payload.
The transfer and withdraw take a payload and the sender public key and a nonce.
The sender account will have the fund removed, and nonce checked/bumped. 
A deposit has no sender account, since it comes from the L1 so it's only a payload.

`fn transfer(&mut self, sender: &PublicKey, transfer: &Transfer, nonce: u64) -> Result<(), TxnErr>`
`fn withdraw(&mut self, withdrawer: &PublicKey, withdraw: &Withdraw, nonce: u64) -> Result<(), TxnErr>`

marijanp (approved):

koxu1996:
**Comment:**

I noticed that `kairos-tx` was [moved](https://github.com/cspr-rad/kairos/pull/98/commits/be0d43d3714636e41fb7f62ea8e3ae61d63da3d8) to the prover directory. It should not be there, as it is a common module used by other root modules (cli, crypto, server) and is not specifically related to the prover itself.

As clarified with @marijanp, this change was implemented to resolve an issue with Nix sourcing. I am really unhappy about such changes. Nix should be an extra feature, not something that compromises the quality of our codebase.

kairos-prover/default.nix
```diff
@@ -44,6 +45,10 @@
             ./rust-std-Cargo.lock
           ];
         };
+
+        # Proving in CI is disabled because it takes too long.
+        # Proving is a test of risc0, not kairos anyway.
+        RISC0_DEV_MODE = 1;
```
koxu1996, line 51: **Question:** Does it also affect our Nix-based production setup? Maybe it should be configured inside CI, not in default Nix config.

kairos-prover/kairos-prover-risc0-server/src/main.rs
```diff
@@ -0,0 +1,210 @@
...
+) -> Result<(ProofOutputs, Receipt), String> {
+    let timestamp = Instant::now();
+
+    let env = ExecutorEnv::builder()
+        .write(&proof_inputs)
+        .map_err(|e| format!("Error in ExecutorEnv builder write: {e}"))?
+        .build()
+        .map_err(|e| format!("Error in ExecutorEnv builder build: {e}"))?;
+
+    let receipt = risc0_zkvm::default_prover()
```
koxu1996, line 74: [Default prover](https://docs.rs/risc0-zkvm/latest/risc0_zkvm/fn.default_prover.html) for non-dev environment is Bonsai. I think we should prefer generating proofs locally, either with `r0vm` subprocess or in-process.

**Request:** Please use explicit type of prover.

kairos-prover/kairos-circuit-logic/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+    pub trie_snapshot: Snapshot<Account>,
+}
+
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Debug, Clone, Eq, PartialEq, PartialOrd, Ord)]
+pub struct ProofOutputs {
+    pub pre_batch_trie_root: TrieRoot<NodeHash>,
+    pub post_batch_trie_root: TrieRoot<NodeHash>,
+    /// TODO consider replacing with a count and hash of the processed deposits
+    pub deposits: Box<[L1Deposit]>,
```
koxu1996, line 32: I think `L1Deposit` should have additional `event_id` field, to uniquely identify deposit. Then proof outputs could contain list of deposit IDs, or even a single number: maximum deposit event ID.

koxu1996:
> TODO (not necessarily before merge):
> - ...
> - get asn1 compiling in risc0-guest for signature verification.

If we want (I think we should) to verify signatures inside ZK VM, then I am not sure why we have duplicate types, e.g. `KairosTransaction` and `Transaction`, but we can resolve this post PR.

koxu1996 (changes requested):
@Avi-D-coder Great work, especially trie integration!

I am requesting one change - [use explicit prover](https://github.com/cspr-rad/kairos/pull/98#discussion_r1624121823) to avoid Bonsai, then I would be happy to approve this PR.

kairos-prover/default.nix
koxu1996: Fixed in https://github.com/cspr-rad/kairos/pull/113.

kairos-prover/kairos-circuit-logic/src/lib.rs
jonas089: We could just make this a Vec of Strings and use hex::encode(deposit.hash())

kairos-prover/kairos-prover-risc0-server/src/main.rs
```diff
@@ -0,0 +1,210 @@
...
+    tracing::info!("Proved batch: {}s", timestamp.elapsed().as_secs_f64());
+
+    let timestamp = Instant::now();
+
+    receipt
+        .verify(PROVE_BATCH_ID)
+        .map_err(|e| format!("Error in risc0_zkvm verify: {e}"))?;
+
+    tracing::info!("Verified batch: {}s", timestamp.elapsed().as_secs_f64());
+
```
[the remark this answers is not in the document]
Avi-D-coder: I'd like to keep it there as a sanity check.
I don't necessarily trust that risc0 will always behave as expected and generate a valid proofs.

This running `verify` only adds about 0.02 seconds to the response time.

kairos-prover/kairos-prover-risc0-server/src/main.rs
jonas089: For the demo it's not that important, it's not hard to fix though, I tend to agree that proving locally is generally better.

kairos-circuit-logic/src/account_trie.rs
```diff
@@ -0,0 +1,349 @@
...
+                pre_batch_trie_root,
+                post_batch_trie_root,
+                deposits: _,
+                withdrawals: _,
+            } = proving_hook(proof_inputs).expect("Failed to prove execution");
+
+            assert_eq!(pre_batch_trie_root, prior_root_hash);
+            assert_eq!(post_batch_trie_root, new_root_hash);
+            prior_root_hash = new_root_hash;
+        }
```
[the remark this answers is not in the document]
Avi-D-coder: commit updates the db.
Not sure what error you are hitting.

I don't get the error and CI passes.
Let's debug on your machine after our meeting.

kairos-prover/kairos-prover-risc0-server/src/main.rs
Avi-D-coder: Fixed
We were using local provers because BONSAI_API_URL and BONSAI_API_KEY were not set, but I agree picking the prover explicitly is better.

kairos-prover/kairos-circuit-logic/src/lib.rs
Avi-D-coder: This is something different.
By hashing here I mean hashing over all the deposits in the batch.
That gives a single hash to serialize, but comes at the cost of an additional sha256 in and outside the vm.

Not sure if it's worth it, but it's a possible future optimization.

kairos-prover/kairos-prover-risc0-server/src/main.rs
koxu1996: I misinterpreted docs, Bonsai was not used then. Thanks for fix!

koxu1996: Fixed in https://github.com/cspr-rad/kairos/pull/98/commits/ee016937731f06a4a1312c8e8dc399e2ee04d8ea.

koxu1996 (approved):

kairos-prover/kairos-circuit-logic/src/transactions.rs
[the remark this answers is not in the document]
Avi-D-coder: Random samples are part of property testing.
In proptest or quickcheck you write code to generate pseudo random instances of your type and then test your code works as expected.

This is better than fixed testing because you cover more of your functions domain.

If your code fails the your test will be run again with a semantically smaller input. This is called shrinking. It allows you to find a minimal example that breaks your assumptions.

jonas089 (approved):
lgtm

marijanp (approved):

PR #105, opened 2024-05-24
Integrate casper event toolkit: use upstream casper-types crate

marijanp:


marijanp:
incorporated these changes based on main in #108 i.e. dismissed in favor of #108

PR #107, opened 2024-05-28
flake: add ci binary cache

marijanp:
- The url and public key allow us to substitute from our CI's nix store
- After discussions with Mark, we figured that the quickest way to accomplish this is by me providing a subdomain


koxu1996 (approved):
I was worried about security, but according to call with @marijanp and @Avi-D-coder, `casper-cache.marijan.pro` is pointing to Hetzner node owned by Casper.

PR #108, opened 2024-05-29
Normalize casper-client, -types and -hashing version

marijanp:


Avi-D-coder (approved):

flake.nix
```diff
@@ -74,6 +74,11 @@
             '';
           };
 
+          # The casper-client's build.rs requires git at buildtime
+          git-mock = pkgs.writeShellScriptBin "git" ''
+            echo "got ya"
```
koxu1996, line 79: ?

marijanp: https://github.com/casper-ecosystem/casper-client-rs/blob/6b61f160a536fafbfb541b1ecb20d05a8e666fd3/build.rs#L7-L12
The comment above explains, maybe I should add the link to the comment. I don't want to make git a build time dependency.

koxu1996 (changes requested):
Not sure what happened here:
1. Why all casper packages are updated to `dev`/`main` branches? I think we decided to postpone Casper 2.0 integration.
2. Why patch is introduced? There should be no conflict in using casper packages, as long they are relying on the same major version.

flake.nix
koxu1996: Is `echo "got ya"` necessary?

marijanp: It can be anything, do you have any preference? I can make it simply exit with a success status. The return value is not used in any significant way.

marijanp:
I gave an explanation in last weeks sync call. The problem is that casper related crates pin different versions of casper-types.

flake.nix
koxu1996: Ah okay, got it.

PR #110, opened 2024-05-30
 integrate casper event toolkit: add unprocessed deposits to batch on deploy

marijanp:
Things done:
- implemented and for a kairos-test-utils/cctl helper that deploys a given contract
  - added a basic test
  - adjusted the nixos module for cctl
- Added a contract hash argument to the kairos-cli deposit command
- Added a known deposit deploys set that is used to keep track of deposit deploys routed through our L2
- Updated the EventManager initialization to use retry and to initialize the next_deposit_index with the current on-chain state
- Added a SSE trigger to the L1Sync service that fetches new events when a notification for a known deploy is received


koxu1996:
@marijanp We should refactor it together once #96 gets merged.

PR #115, opened 2024-06-05
Profile risc0 and use AccountTrie in server

Avi-D-coder:
Merge https://github.com/cspr-rad/kairos/pull/94 before this PR.

## Changes

We now use the same code in the circuit and server.
We now precheck transactions before attempting to include them in a batch.
We now rely on the security of sha(public_key) instead of also checking the public key matches the one used to create an account. This made the circuit a bit more efficient (see 3526834d695b8842d142422d5efd3f869f275a17).

## Profiling
Our circuit is already pretty optimized unfortunately.

60% of cycles are spent calculating the merkle root, 30% on deserializing the data into risc0.
Maybe we can use the older risc0 custom sterilization instead of serde.
I can align the data a bit better, maybe I can optimize `calc_subtree_hash` a bit, but not much.

```
Flat	Flat%	Sum%	Cum	Cum%	Name	Inlined?
2229336	24.11%	24.11%	2229336	24.11%	sha2::sha256::compress256	
1405433	15.20%	39.30%	1405433	15.20%	memcpy	
803798	8.69%	48.00%	2954603	31.95%	<kairos_trie::stored::merkle::Snapshot<V> as kairos_trie::stored::Store<V>>::calc_subtree_hash	
788440	8.53%	56.52%	5509723	59.58%	kairos_trie::transaction::Transaction<S,V>::calc_root_hash_node	
149716	1.62%	58.14%	149716	1.62%	memset	
133000	1.44%	59.58%	443959	4.80%	kairos_trie::transaction::nodes::Leaf<V>::hash_leaf	
0	0.00%	59.58%	5509723	59.58%	main	
0	0.00%	59.58%	5509723	59.58%	__start
```


jonas089 (approved):
lgtm

kairos-prover/kairos-circuit-logic/src/account_trie.rs
```diff
@@ -110,9 +114,10 @@ impl<S: Store<Account>> AccountTrie<S> {
             )
         })?;
 
-        if sender_account.pubkey != *sender {
-            return Err(("hash collision detected on sender account").into());
-        }
+        // SECURITY ASUMPTION: see Account docs
+        // if sender_account.public_key != *sender {
+        //     return Err(("hash collision detected on sender account").into());
+        // }
```
marijanp, line 120: Can we remove this?

marijanp (approved):
After completing the review I assume that the SECURITY ASSUMPTION comments are intended, so lgtm

kairos-prover/kairos-circuit-logic/src/account_trie.rs
Avi-D-coder: I just left that anywhere I made the assumption.

kairos-tx/src/error.rs
```diff
@@ -35,6 +35,7 @@ mod error {
     use super::*;
     use core::fmt::{Debug, Display};
 
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
Avi-D-coder: Hopefully not, but maybe.
no_std `Error` should be in 1.80 I think.

PR #122, opened 2024-06-13
Use workspace deps for casper related crates

marijanp:


Avi-D-coder (approved):

Cargo.toml
```diff
@@ -17,3 +17,11 @@ name = "kairos"
...
 edition = "2021"
 license = "MIT OR Apache-2.0"
+
+[workspace.dependencies]
+casper-client = "2.0"
+casper-types = "4.0"
+casper-client-types = { package = "casper-types", version = "3.0" }
+casper-hashing = "3.0"
+casper-event-standard = "0.5"
+casper-event-toolkit = { git = "https://github.com/koxu1996/casper-event-toolkit", branch = "test" }
```
koxu1996, line 27:
```suggestion
casper-event-toolkit = { git = "https://github.com/koxu1996/casper-event-toolkit", version = "0.1.3" }
```

kairos-cli/Cargo.toml
```diff
@@ -18,8 +18,8 @@ all-tests = ["cctl-tests"]
 cctl-tests = []
 
 [dependencies]
-casper-client = "2"
-casper-types = { version = "3", features = ["std"] } # TODO: Change `std` -> `std-fs-io` in the future version.
```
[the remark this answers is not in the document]
marijanp: as you can see i removed v3, v4 is not compatible with casper-client 2

kairos-contracts/demo-contract/contract-utils/Cargo.toml
```diff
@@ -4,6 +4,11 @@ version.workspace = true
...
+[lib]
+test = false
+bench = false
+doctest = false
+
 [dependencies]
-casper-types = { version = "3", default-features = false }
-casper-event-standard = { version = "0.5.0", default-features = false }
+casper-types.workspace = true
+casper-event-standard = { version = "0.5", default-features = false }
```
koxu1996, line 14: This should use common dependency as well:

```suggestion
casper-event-standard = { workspace = true, default-features = false }
```

demo-contract-tests/Cargo.toml
```diff
@@ -8,6 +8,6 @@ license.workspace = true
 casper-engine-test-support = { version = "7.0.0", default-features = false }
 casper-execution-engine = { version = "7.0.0", default-features = false }
 casper-contract = { version = "4.0.0", default-features = false }
```
koxu1996, line 10: This should use common dependency too:

```suggestion
casper-contract = { workspace = true, default-features = false }
```

kairos-cli/Cargo.toml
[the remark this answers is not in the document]
koxu1996: Okay, I see that https://github.com/cspr-rad/kairos/pull/94 downgraded types recently - I missed it.

koxu1996 (approved):
@marijanp I analyzed usage of `casper-*` dependencies and found 2 missed cases - suggestions added.

---

Extra notes from my review:

### Old casper-types

We were using old casper types in 3 places:

```
kairos-cli/Cargo.toml:22:casper-types = { version = "3", features = ["std"] } # TODO: Change `std` -> `std-fs-io` in the future version.
kairos-server/Cargo.toml:50:casper-types = "3"
kairos-contracts/demo-contract/contract-utils/Cargo.toml:8:casper-types = { version = "3", default-features = false }
```

This PR replaced first two with the common `casper-client-types = { package = "casper-types", version = "3.0" }`, which is good, just a bit of renaming. However, contract-utils was upgraded to types `4.0`, which is okay too, because we were using old types without valid reason.

So now we are using old types in 2 places: `kairos-cli` and `kairos-server`. I will do some cleanup later, as we can reduce it to single usage.

### `casper-*` dependencies with a single usage

Test crates:

```
demo-contract-tests/Cargo.toml:8:casper-engine-test-support = { version = "7.0.0", default-features = false }
demo-contract-tests/Cargo.toml:9:casper-execution-engine = { version = "7.0.0", default-features = false }
```

CES toolkit:

```
kairos-server/Cargo.toml:43:casper-event-toolkit = { git = "https://github.com/koxu1996/casper-event-toolkit.git", version = "0.1.3" }
```

Cargo.toml
```diff
@@ -18,3 +18,11 @@ name = "kairos"
 version = "0.1.0"
 edition = "2021"
 license = "MIT OR Apache-2.0"
+
+[workspace.dependencies]
+casper-client = "2.0"
+casper-types = "4.0"
+casper-client-types = { package = "casper-types", version = "3.0" }
+casper-client-hashing = { package = "casper-hashing", version = "2.0" }
+casper-contract = "4.0"
```
koxu1996, line 27: Please add missing `default-features = false`, as features as additive and that's causing build to fail:

```suggestion
casper-contract = { version = "4.0", default-features = false }
```

koxu1996 (approved):

PR #124, opened 2024-06-13
Data availability

Rom3dius:
Stores withdrawals, transfers and deposits in a single postgresql table. Allows to query for a maximum amount of 500 transactions, using the following filters:
- sender
- recipient
- min amount
- max amount
- min timestamp
- max timestamp

Timestamp is created right before insert.

Testing needs work, tests currently fail because test setup is synchronous but pool setup needs async runtime. And of course the missing database. This is the current testing connection string: `postgres://kairos:kairos@localhost/kairos`


kairos-data/src/transaction.rs
```diff
@@ -0,0 +1,137 @@
...
+#[diesel(table_name = transactions)]
+#[diesel(check_for_backend(diesel::pg::Pg))]
+pub struct Transaction {
+    pub timestamp: NaiveDateTime,
+    pub public_key: String,
+    pub nonce: Option<i64>,
+    pub trx: i16,
+    pub amount: i64,
+    pub recipient: Option<String>,
+}
```
Avi-D-coder, line 24: Can we normalize this table.
That would remove the need for the enum tag `trx`.

kairos-data/src/transaction.rs
```diff
@@ -0,0 +1,137 @@
...
+        }
+    }
+}
+
+impl From<Signed<Withdraw>> for Transaction {
+    fn from(signed_withdraw: Signed<Withdraw>) -> Self {
+        Transaction {
+            timestamp: Utc::now().naive_utc(),
+            public_key: hex::encode(&signed_withdraw.public_key),
+            nonce: Some(signed_withdraw.nonce as i64),
```
Avi-D-coder, line 118: `u64 as i64` may not preserve the value.
You could use `try_into` and error, but it's not ideal.

Try using `NUMERIC`.
https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-NUMERIC-DECIMAL
https://github.com/launchbadge/sqlx/discussions/2977#discussioncomment-8428838

kairos-server/src/routes/transfer.rs
```diff
@@ -35,12 +36,14 @@ pub async fn transfer_handler(
 
     tracing::info!("queuing transaction for trie update");
 
+    let transfer = KairosTransaction::Transfer(Signed {
+        public_key,
+        nonce,
+        transaction: transfer,
+    });
+    db::insert(state.pool.clone(), transfer.clone()).await?;
```
Avi-D-coder, line 44: Let's put the db access behind a feature flag so that we don't need postgres to run simple tests.

kairos-data/src/transaction.rs
Rom3dius: We could, but it adds needless complication querying with little benefit. Nothing needs to be done from the API side for the crate to figure out which enum it should be.

kairos-data/src/transaction.rs
Rom3dius: NUMERIC in Diesel is represented by BigDecimal. I swapped it over to BigDecimal

kairos-server/src/routes/transfer.rs
Rom3dius: Done.

Cargo.toml
```diff
@@ -9,7 +9,7 @@ members = [
   "kairos-test-utils",
   "kairos-tx",
   "kairos-prover/kairos-circuit-logic",
-  "demo-contract-tests",
+  "demo-contract-tests", "kairos-data",
```
marijanp, line 12: Nit: add newline here

kairos-data/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+
+pub use deadpool_diesel::postgres::Pool;
+pub use diesel::{insert_into, prelude};
+
+pub mod errors;
+pub mod schema;
+pub mod transaction;
+
+#[cfg(feature="migrations")]
+pub const MIGRATIONS: EmbeddedMigrations = embed_migrations!();
```
marijanp, line 16: I would prefer the migrations directory to be parametrzed and configurable when I call `run`. It could also be an option and default to whatever the default in the `Cargo.toml` is.

kairos-data/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+        .build()
+        .unwrap();
+    debug!("Initialized connection pool.");
+    #[cfg(feature="migrations")]
+    run_migrations(&conn_pool).await;
+    let conn = conn_pool.get().await?;
+    let result = conn.interact(|conn| {
+        let query = select("Hello world!".into_sql::<Text>());
+        query.get_result::<String>(conn)
+    }).await??;
```
marijanp, line 32: Nit: Add comment why we do this query.

kairos-server/tests/transactions.rs
```diff
@@ -48,7 +48,7 @@ fn new_test_app_with_casper_node(casper_node_url: &Url) -> TestServer {
         batch_state_manager: BatchStateManager::new_empty(),
         server_config: ServerConfig {
             socket_addr: "0.0.0.0:0".parse().unwrap(),
-            casper_rpc: casper_node_url.clone(),
+            casper_rpc: casper_node_url.clone(),
```
marijanp, line 51: Remove these spaces

kairos-server/src/state.rs
```diff
@@ -10,12 +10,16 @@ use crate::config::ServerConfig;
 use kairos_circuit_logic::transactions::KairosTransaction;
 use kairos_trie::{stored::memory_db::MemoryDb, NodeHash, TrieRoot};
 
+#[cfg(feature="database")]
+use kairos_data::Pool;
+
 pub type ServerState = Arc<ServerStateInner>;
 
-#[derive(Debug)]
```
marijanp, line 15: Why was this trait removed?

kairos-server/Cargo.toml
```diff
@@ -16,6 +16,7 @@ default = ["cctl-tests", "deposit-mock"]
 all-tests = ["cctl-tests", "deposit-mock"]
 cctl-tests = []
 deposit-mock = []
+database = ["kairos-data"]
```
marijanp, line 19: The feature should probably be added to `all-tests`

kairos-server/src/routes/fetch.rs
```diff
@@ -0,0 +1,43 @@
...
+    sender: Option<String>,
+    min_timestamp: Option<String>,
+    max_timestamp: Option<String>,
+    min_amount: Option<i64>,
+    max_amount: Option<i64>,
+    recipient: Option<String>,
+}
+
+#[derive(TypedPath)]
+#[typed_path("/api/v1/query_transactions")]
```
marijanp, line 21: I would suggest naming the route `api/v1/tranasactions`

kairos-server/src/routes/fetch.rs
```diff
@@ -0,0 +1,43 @@
...
+    State(state): State<ServerState>,
+    Json(payload): Json<QueryTransactionsPayload>,
+) -> Result<Json<Vec<Transaction>>, AppErr> {
+    let filter = TransactionFilter {
+        sender: payload.sender,
+        min_timestamp: payload.min_timestamp.and_then(|ts| NaiveDateTime::parse_from_str(&ts, "%Y-%m-%d %H:%M:%S").ok()),
+        max_timestamp: payload.max_timestamp.and_then(|ts| NaiveDateTime::parse_from_str(&ts, "%Y-%m-%d %H:%M:%S").ok()),
+        min_amount: payload.min_amount,
+        max_amount: payload.max_amount,
+        recipient: payload.recipient,
```
marijanp, line 36: Seems like a good opportunity to introduce a `From` implementation for the payload.

kairos-server/src/routes/fetch.rs
```diff
@@ -0,0 +1,43 @@
...
+    };
+
+    let transactions = get(state.pool.clone(), filter)
```
marijanp, line 39: If we make `get` take a reference we could avoid cloning here, or do we need to take ownership?

Cargo.toml
Rom3dius: No idea why I left it like that, will fix

kairos-data/migrations/2024-06-13-092829_transactions/down.sql
```diff
@@ -0,0 +1,2 @@
+-- This file should undo anything in `up.sql`
+DROP TABLE transactions;
```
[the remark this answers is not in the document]
Rom3dius: Looks like my editor doesnt do it automatically for .sql files

kairos-data/migrations/2024-06-13-092829_transactions/up.sql
```diff
@@ -0,0 +1,11 @@
+-- Your SQL goes here
+-- CREATE TYPE transaction AS ENUM ('Deposit', 'Transfer', 'Withdrawal');
```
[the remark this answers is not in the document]
Rom3dius: Sure

kairos-data/src/lib.rs
Rom3dius: The migrations directory can be configured from diesel.toml

kairos-server/src/state.rs
Rom3dius: Pool doesn't implement Debug

kairos-server/src/routes/fetch.rs
Rom3dius: Nah we don't need to clone, fixed

kairos-server/src/routes/fetch.rs
Rom3dius: No need for an abstraction, there's not gonna be multiple API endpoints that use this payload struct and have to make a filter out of it.

kairos-server/src/routes/fetch.rs
Rom3dius: Sure

kairos-server/Cargo.toml
Rom3dius: Sure

kairos-data/migrations/2024-06-13-092829_transactions/up.sql
```diff
@@ -0,0 +1,11 @@
...
+CREATE TABLE transactions (
+    "timestamp" timestamp DEFAULT CURRENT_TIMESTAMP,
+    public_key varchar NOT NULL,
+    nonce numeric,
+    trx smallint NOT NULL,
```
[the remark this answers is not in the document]
Rom3dius: Added

kairos-data/src/transaction.rs
```diff
@@ -0,0 +1,143 @@
...
+                .select(Transactions::as_select())
+                .limit(500)
+                .load::<Transactions>(conn)
+        })
+        .await??;
+    Ok(res)
+}
+
+pub async fn insert(
+    pool: crate::Pool,
```
marijanp, line 81: Use a reference here too.

marijanp: In general, I think the right approach is to take references for complex parameter datatypes until ownership is strictly necessary. That way we can avoid clones.

marijanp (approved):
Just the references for insert and query is what I would like to see to avoid clones.
Tests for the `transactions` endpoint testing the filter should/could be added in a successive PR.
Otherwise LGTM

koxu1996 (approved):
LGTM. One remark: I think we should store transactions in `kairos-tx` format.

kairos-data/src/transaction.rs
jonas089: So `trx` is either 0,1,2 mapped to (deposit, transfer, withdrawal)? lgtm.

jonas089 (approved):
lgtm

koxu1996:
Superseded by https://github.com/cspr-rad/kairos/pull/147.

PR #120, opened 2024-06-10
Feature/contract submit batch

jonas089:
Objective of this PR:
Add Batch verifier entry point to the demo contract

Todo before merge:

- remove temp-e2e-tests
- resolve merge conflicts
- discuss and implement sliding window pointer update

Optional (but likely appreciated):

- don't depend on my github
```
replace 
{ git = "https://github.com/jonas089/casper-node", branch = "kairos-demo-chainspec", default-features=false }
with
{ git = "https://github.com/cspr-rad/casper-node", branch = "kairos-demo-chainspec", default-features=false }
in the contract integration tests - the 
```

Post demo todos (irrelevant to the demo):

- more efficient serialization (for example Borsh), purge serde-json-wasm
- verify deposit hashs when verifying batches, BEFORE updating the root
- potentially merge the nix build commands back into one 
```
nix build .#kairos-demo-contracts
nix build .#kairos-session-code
```
This is for Marijan to decide, I just did what I thought was convenient to get it to work.

PR #125, opened 2024-06-14
nixos/cctl: use exponential backoff, remove unused code

marijanp:
- Added a chainspec nixos option
- Updated the e2e to use a chainspec that has bigger deploy, block and arg length sizes: https://github.com/jonas089/casper-node/blob/kairos-demo-chainspec/resources%2Flocal%2Fconfig.toml
- added an exponential backoff when waiting for genesis


Avi-D-coder (approved):

Avi-D-coder:
I don't think this is working.
I am setting the chainspec latest commits of but the nixos vm test still fails with the same deploy size error. in https://github.com/cspr-rad/kairos/pull/120

The error still states the default max_deploy size not the changed one.

koxu1996:
There are 2 **different** configs file used to configure Casper node:
 - config.toml
 - chainspec.toml

This PR pulls _config_ and incorrectly use it as a _chainspec_. To increase max deploy size you need to customize both - see https://github.com/cspr-rad/kairos/pull/120#issuecomment-2167939187.

Instead of pulling those files from custom node fork, I would rather use [official protocol release](https://github.com/casper-network/casper-protocol-release/tree/casper-1.5.6/config) with 2 `sed` commands, or patch file.

Avi-D-coder:
> There are 2 **different** configs file used to configure Casper node:
> 
>     * config.toml
> 
>     * chainspec.toml
> 
> 
> This PR pulls _config_ and incorrectly use it as a _chainspec_. To increase max deploy size you need to customize both - see [#120 (comment)](https://github.com/cspr-rad/kairos/pull/120#issuecomment-2167939187).
> 
> Instead of pulling those files from custom node fork, I would rather use [official protocol release](https://github.com/casper-network/casper-protocol-release/tree/casper-1.5.6/config) with 2 `sed` commands, or patch file.

I switched it to `chainspec.toml` in https://github.com/cspr-rad/kairos/pull/120/commits/3f4b56f33dff42b8292fa54a89358f61af48e55b, I did not include the changed config. Maybe that's why the error is still happening, but the error still states `kairos: output: deploy size of 1465741 bytes exceeds limit of 1048576`, `1048576` is the unmodified chainspec.toml value, so don't think the chainspec flag is doing anything.

marijanp:
@Avi-D-coder the problem was also due to an incorrect argument format for the cctl command. It expects the arguments to be passed as `config=<path> chainspec=<path>` it should be fixed now I've verified the generated files manually, please try again.

marijanp:
@koxu1996 for now the goal was to get a working as simple as possible. I would also prefer to introduce nixos options where a user can say
```
services.cctl.maxBodyBytes = <num-bytes>;
services.cctl.maxBlockSize = <size>;
```

Avi-D-coder (approved):

Avi-D-coder:
I prefer the whole file style, I would rather not have to match the seds to the changes made in the upstream files we use in the rust contract tests.

https://github.com/cspr-rad/kairos/blob/2dd35183d7ad0112cc55500d429a67c12a11f5ef/nixos/tests/end-to-end.nix#L29C1-L39C9

koxu1996 (approved):

koxu1996:
@Avi-D-coder What about using [official config files](https://github.com/casper-network/casper-protocol-release/tree/casper-1.5.6/config) instead of those from `cspr-rad` fork?

koxu1996:
@marijanp They are **almost** the same. I approved PR, but using official protocol releases would be better.

Avi-D-coder (approved):

PR #127, opened 2024-06-17
call proving server from kairos server

Avi-D-coder:
This pr just enables proving batches, and adds some test utils.


marijanp (approved):

asladeofgreen (approved):
Seems consistent.

PR #132, opened 2024-06-22
Fix deposit type emitted from contract.

Avi-D-coder:
This will be switched to a pr against main after `feature/contract-submit-batch` is merged.


Avi-D-coder:
This pr is blocked on the cli implementation

kairos-session-code/deposit-session/src/main.rs
```diff
@@ -30,6 +31,7 @@ pub extern "C" fn call() {
         "deposit",
         runtime_args! {
             "temp_purse" => temp_purse,
+            "recipient" => public_key,
```
jonas089, line 34: Should this really be a Casper PublicKey? Shouldn't this be a Kairos PublicKey?

jonas089: Other than that lgtm.

Avi-D-coder: Kairos PublicKey is just Vec<u8> which would need CLType::Any, so I convert to bytes in the contract instead.

Avi-D-coder:
The nixos test is still broken, but please review.

marijanp (approved):

PR #139, opened 2024-06-25
Fixup deposit counter type

marijanp:
`u64` is not compatible with the casper-event-standard: https://github.com/make-software/casper-event-standard/blob/ad942762a099fc29c0df4cbd3604da8201704195/casper-event-standard/src/contract.rs#L21
We us the last_processed_deposit_index for our deposit window, and we could end up with a much bigger index there than the events length

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1203,7 +1203,6 @@ name = "contract-utils"
 version = "0.1.0"
 dependencies = [
  "casper-event-standard",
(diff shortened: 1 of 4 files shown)


Avi-D-coder (approved):
Yikes, worried about this. We will definitely have to fix this post demo for the sliding window.

quinn-dougherty (approved):

Avi-D-coder (approved):

PR #143, opened 2024-07-08
Reduce WASM size to fit into production chainspec

koxu1996:
## Context

In Casper v1.5.6 [maximum contract size](https://github.com/casper-network/casper-protocol-release/blob/ce0c03e4b90a2abc0066a2dc13da38d667381423/config/chainspec.toml#L133-L134) is 1 MiB (1,048,576 bytes), which we fairly exceed with demo contract (1.4 MiB), therefore making Kairos incompatible with production network.

## Can we reduce it?

I was suspecting that Risc0 verifier might be the biggest part of final WASM, so I did some tests on [bare minimum contract](https://gist.github.com/koxu1996/9a01db96c632b41268ed0138dd75d01f). However, bare Risc0 verifier (+ bincode2 deserialization for receipt) is less than 20 KB :space_invader:. Casper API calls are also quite small (few kilobytes), so it means we should be able to get way less than 1 MiB.

## First easy optimization

Simple idea: we can enable more aggressive optimization in Rust complier and WASM optimizer.

I run benchmark for WASM produced by `nix build -L --no-link --show-trace .#packages.x86_64-linux.kairos-contracts`:

- Before PR: **1.4MiB** (1,374,581 bytes)
- Setting Rust `o='s'`: **1004KiB** (1,025,823 bytes)
- Setting Rust `o='z'`: **904KiB** (921,813 bytes)
- Setting WASM `o='s'`: **1.3MiB** (1,322,590 bytes)
- Setting WASM `o='z'`: **1.3MiB** (1,320,589 bytes)
- Setting Rust `o='s'` and WASM `o='s'`: **964KiB** (985,116 bytes)
- Setting Rust `o='s'` and WASM `o='z'`: **964KiB** (983,713 bytes)
- Setting Rust `o='z'` and WASM `o='s'`: **868KiB** (886,560 bytes)
- Setting Rust `o='z'` and WASM `o='z'`: **868KiB** (884,798 bytes) :arrow_left: This PR.

After those optimizations we are using 84.77% of production WASM limit :relieved:. Still much, but it was low hanging fruit.

### DIFF
--- a/demo-contract-tests/tests/test_fixture/mod.rs
+++ b/demo-contract-tests/tests/test_fixture/mod.rs
@@ -201,3 +201,27 @@ pub fn create_funded_account_for_secret_key_bytes(
     builder.exec(transfer).expect_success().commit();
     account_public_key
 }
+
+/// Returns an [`ExecuteRequest`] that will call a stored contract by hash.
+/// NOTE: This utiliity is a custom version of `ExecuteRequest::contract_call_by_hash`
+/// that allows to specify custom payment.
+pub fn contract_call_by_hash(
+    sender: AccountHash,
+    contract_hash: ContractHash,
+    entry_point: &str,
+    args: RuntimeArgs,
+    payment: U512,
+) -> ExecuteRequestBuilder {
+    let mut rng = rand::thread_rng();
+    let deploy_hash = rng.gen();
+
+    let deploy = DeployItemBuilder::new()
+        .with_address(sender)
+        .with_stored_session_hash(contract_hash, entry_point, args)
+        .with_empty_payment_bytes(runtime_args! { ARG_AMOUNT => payment, })
+        .with_authorization_keys(&[sender])
+        .with_deploy_hash(deploy_hash)
+        .build();
+
+    ExecuteRequestBuilder::new().push_deploy(deploy)
+}
--- a/demo-contract-tests/tests/test_fixture/mod.rs
+++ b/demo-contract-tests/tests/test_fixture/mod.rs
@@ -1,7 +1,7 @@
 mod wasm_helper;
 
 use casper_engine_test_support::{
-    ExecuteRequestBuilder, WasmTestBuilder, ARG_AMOUNT, DEFAULT_ACCOUNT_ADDR,
+    DeployItemBuilder, ExecuteRequestBuilder, WasmTestBuilder, ARG_AMOUNT, DEFAULT_ACCOUNT_ADDR,
     DEFAULT_ACCOUNT_INITIAL_BALANCE,
 };
 use casper_execution_engine::storage::global_state::in_memory::InMemoryGlobalState;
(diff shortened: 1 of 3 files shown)


Avi-D-coder (approved):

Avi-D-coder:
i had experimented with this and had not had much luck. If you can get it to pass the contract tests I'm all for it.

marijanp:
Looks good, any idea why we are running out of gas in the tests?

marijanp (approved):

jonas089 (approved):
LGTM. Nice optimization.

koxu1996:
@Avi-D-coder @marijanp Contract calls in test engine [are using the default payment](https://github.com/casper-network/casper-node/blob/9f3995853204a18f17de9c022233d22aa14b9c37/execution_engine_testing/test_support/src/execute_request_builder.rs#L124), which is [2500 CSPR](https://github.com/casper-network/casper-node/blob/9f3995853204a18f17de9c022233d22aa14b9c37/execution_engine_testing/test_support/src/lib.rs#L131).

Since we optimized code for size, it might be **more expensive** to execute. I just measured it and previously the cost was 1043 CSPR, and after full size optimization it is 2756 CSPR :fearful:. For example just by setting Rust `o='s'` we get 1004KiB (still fits in production chainspec) and execution takes 1588 CSPR.

**We should find a good balance between WASM size and cost of execution, but I think compatibility with the production chainspec is a priority here.**

---

Code used for testing:
```rs
        let cost = self
            .builder
            .get_last_exec_results()
            .unwrap()
            .get(0)
            .unwrap()
            .cost();
        panic!("Cost of execution: {:?}", cost);
```

Avi-D-coder:
@koxu1996 try build-std https://doc.rust-lang.org/cargo/reference/unstable.html

koxu1996:
> > Risc0 verifier (+ bincode2 deserialization for receipt)
> 
> Make sure you use the code when judging the size, only code paths you use and ones that could be called by dyn trait make it into the optimized binary. Our contract also has std from risc0 transitive dependencies linked in.

Yup, I made unsuccessful verification trigger panic, so Risc0 code was definitely used there.

> @koxu1996 try build-std https://doc.rust-lang.org/cargo/reference/unstable.html

On the one hand Risc0 has a bug, where `std` is introduced in one of old dependencies (`downcast-rs` in `rrs-lib`). However, it gets optimized out, since it's not used in our code. Overall, `std` library is not present in Kairos contracts, so there is no need for `build-std`.

Avi-D-coder:
Hmm, that seems odd.
What is the main contributor to our binary size then?

koxu1996:
I don't like `serde_json`... but this needs more investigation, as we pull quite a lot of dependencies.

Avi-D-coder:
Try borsh we could not get bincode working in the wasm contract, float related issue.

PR #146, opened 2024-07-09
Add get-nonce endpoint

marijanp:
See comment in the e2e why we cannot reliably test this yet


kairos-prover/kairos-circuit-logic/src/account_trie.rs
```diff
@@ -199,6 +199,16 @@ impl<S: Store<Account>> AccountTrie<S> {
 
         Ok(())
     }
+
+    /// Returns the nonce for an accounts public key if it's known, returns an error if unknown.
+    pub fn get_nonce_for(&self, account: &PublicKey) -> Result<u64, TxnErr> {
+        let [account_hash] = hash_buffers([account]);
+
+        let account = self.txn.get(&account_hash)?;
```
Avi-D-coder, line 207: Using get adds dependencies to the transaction's snapshot.
Use `get_exclude_from_txn`
```suggestion
        let account = self.txn.get_exclude_from_txn(&account_hash)?;
```

Avi-D-coder (changes requested):

kairos-prover/kairos-circuit-logic/src/account_trie.rs
marijanp: done

Avi-D-coder (approved):

koxu1996:
I am not sure about that edge case. `nonce` is used to count number of interactions between user and L2, and deposit is not using it, since it is initiated on L1. Should not first transfer use `nonce=0`?

marijanp:
@koxu1996 after successful batch verification of a deposit the account trie will be updated by inserting an account. And that accounts nonce defaults to 0. Is that what you are referring to?

Avi-D-coder:
I think synchronization here is just running get nonce in a loop.
The account is added to the trie as soon as the l1 even hits the l2 server.
It does not wait until batch verification. The proof of the next batch contains the output of a matching deposit and then the pending deposit is complete.

koxu1996:
> after successful batch verification of a deposit the account trie will be updated by inserting an account. And that accounts nonce defaults to 0. Is that what you are referring to?
 
Not really. Every public key you could imagine like 0000..0000 is valid account, and nonce is just number of interactions with L2. Therefore I think that nonce returned for account that does not exist in trie should be 0.

nixos/tests/end-to-end.nix
```diff
@@ -139,6 +139,10 @@ nixosTest {
 
     # transfer
     beneficiary = client.succeed("cat ${clientUsersDirectory}/user-3/public_key_hex")
+    # TODO we currently have to pass the nonce explicitly as we are lacking a way to synchronize
+    # the successful creation of an account in the account trie after a deposit.
+    # After adding the DA and awaiting that the server picked up the new deposit and thus
+    # an according account was added to the trie, we can discard the nonce and get it from the server
```
koxu1996, line 145: Wait mechanism for deposit is necessary indeed, but it does not have to be done with DA.

koxu1996 (approved):

marijanp:
Don't merge yet

marijanp:
ready to merge

PR #148, opened 2024-07-11
Add missing tests for Deploy Notifier (SSE)

### LINKED ISSUE #85: Missing tests in Casper Deploy Notifier
See https://github.com/cspr-rad/kairos/pull/40#pullrequestreview-2029228384.

koxu1996:
Fixes #85.


Avi-D-coder (approved):

marijanp (approved):
Mocking makes it difficult to detect regressions: https://cs-syd.eu/posts/2021-10-22-why-mocking-is-a-bad-idea

koxu1996:
@marijanp Provided tests with mocks are mostly the evidence that edge cases are handled according to my expectations: scenarios like double handshake should never happen. Of course we need integration tests to have higher confidence that code is actually working. If I have time I will experiment with running NCTL for notifier tests.

PR #160, opened 2024-07-23
Built-in customizable deposit WASM

### LINKED ISSUE #159: Client relies on hardcoded path to deposit session
When making deposit `kairos-client` binary is **reading WASM from path set during compile time**:

https://github.com/cspr-rad/kairos/blob/b718d644703aa077cc62653248ff446403002fff/kairos-cli/src/client.rs#L57-L59

This means:

- You cannot move project, as WASM files must stay at previous path.
- You cannot move `kairos-client` binary to other PC.

koxu1996:
Fixes #159 by:
- including WASM bytes during compile time - still relying on previously used env `PATH_TO_SESSION_BINARIES`
- adding **optional** argument that allow specifying custom WASM:
```
$ ./kairos-cli deposit --help

[cut]

  -s, --session-path <PATH>            Path to the custom WASM session code for deposit
```


kairos-cli/build.rs
```diff
@@ -0,0 +1,20 @@
...
+    // Get the output directory set by Cargo.
+    let out_dir = env::var("OUT_DIR").expect("OUT_DIR not set");
+    let source_path = Path::new(&session_binaries_dir).join("deposit-session-optimized.wasm");
+    let dest_path = Path::new(&out_dir).join("deposit-session-optimized.wasm");
+
+    // Copy the file from the source to the destination
+    fs::copy(&source_path, dest_path).expect("Failed to copy WASM file");
+
+    // Print out a message to re-run this script if the source file changes.
+    println!("cargo:rerun-if-changed={}", source_path.display());
```
Avi-D-coder, line 19: NIT: This works, but I am not clear why we need to copy the wasm into `OUT_DIR`. Why not just `include_bytes` from PATH_TO_SESSION_BINARIES directly?

Avi-D-coder (approved):

kairos-cli/build.rs
koxu1996: `include_bytes!(env!("PATH_TO_SESSION_BINARIES"))` would work too. However, I would like to extend build script to automatically discover and optimize WASMs (similar to [this](https://github.com/cspr-rad/kairos/blob/fdcd1f595d0067e37f936046097787c3c2cab666/demo-contract-tests/tests/test_fixture/wasm_helper.rs#L6)) so `PATH_TO_SESSION_BINARIES` can be avoided.

The whole project works great in Nix, but being able to just `cargo build` without bunch of extra env setup would be better.

Avi-D-coder: Sounds good

marijanp (approved):

